PR #11, opened 2016-03-24
Various fixes

Crashfreak:
Most of the fixes here are related to exception handling, but some UI clean up and user management as well.


app/assets/javascripts/app.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * Created by cuthbertm on 3/21/16.
+ */
+toastr.options = {
```
mvexel: I [used toastr for MapRoulette 1](https://github.com/osmlab/maproulette/blob/dd0c0bc3aa2dd2d4bdf86ebd70f1f84cd6271b26/jsx/maproulette.js#L16) because it was the most lightweight thing I could find that does what I needed it to (stylable, supports html content, variable auto dismiss time, sticky option, perhaps a few other minor things). I don't see a reason to change toast message frameworks but just curious if you had looked at any others?

app/assets/javascripts/mapping.js
```diff
@@ -69,16 +28,24 @@ var MRConfig = (function () {
 (function() {
     var map;
     $(document).ready(function() {
-        var road_layer = new L.TileLayer('http://otile{s}.mqcdn.com/tiles/1.0.0/map/{z}/{x}/{y}.png', {
```
mvexel: We could clean this up a little by using [these shortcuts](https://gist.github.com/mourner/1804938). No idea if these ever made it into Leaflet core.

app/assets/javascripts/mapping.js
```diff
@@ -69,16 +28,24 @@ var MRConfig = (function () {
...
+            maxZoom: 18,
+            attribution: "&copy; <a href=\'http://openstreetmap.org\'> OpenStreetMap</a> contributors"
+        }),
+        road_layer = new L.TileLayer('http://otile{s}.mqcdn.com/tiles/1.0.0/map/{z}/{x}/{y}.png', {
             maxZoom: 18,
             subdomains: ['1', '2', '3', '4'],
             attribution: 'Tiles Courtesy of <a href="http://www.mapquest.com/" target="_blank">MapQuest</a>. Map data (c) <a href="http://www.openstreetmap.org/" target="_blank">OpenStreetMap</a> contributors, CC-BY-SA.'
         }),
-        satellite_layer = new L.TileLayer('http://otile{s}.mqcdn.com/tiles/1.0.0/sat/{z}/{x}/{y}.png', {
+        mapquest_layer = new L.TileLayer('http://otile{s}.mqcdn.com/tiles/1.0.0/sat/{z}/{x}/{y}.png', {
```
mvexel: the Mapquest layers allow only for a limited number of transactions (currently 15000) under the free plan. 
We can get free Skobbler / Telenav tiles for sure.
ESRI tiles in MR1 are used under a 'gentleman's agreement'. I'll need to confirm with them if we can continue using those. I'll create a separate ticket to track this.

Crashfreak: That is good to know, currently I was just playing around with it to see what was available and what could be done. I think when we are closer to the finished product we can nail down exactly what this should look like.

app/assets/javascripts/app.js
Crashfreak: I did not, I am not using it very much at the moment, just one minor thing really which I will probably remove anyway. It seemed to work pretty well and easily so I didn't see a reason to change it either.

app/assets/javascripts/mapping.js
Crashfreak: I doesn't look like it is in 0.7.7 at least, but I leveraged the idea which will apply in the next PR.

PR #15, opened 2016-04-01
Enable Travis / Heroku

mvexel:
A few small tweaks to enable an automated CI flow. Also a small fix to the migration (gave an error because of invalid ID on initial creation of the database. 

**Note that you need to stick the database connection string (including user/pass) in an environment variable after this gets merged.** Example: `DATABASE_URL=jdbc:postgresql://osm:osm@localhost:5432/mp_dev`


build.sbt
```diff
@@ -2,7 +2,9 @@ name := "PlayMapRouletteV2"
 
 version := "1.0"
 
-lazy val `playmaproulettev2` = (project in file(".")).enablePlugins(PlayScala)
+lazy val `playmaproulettev2` = (project in file("."))
+  .enablePlugins(PlayScala)
+  .enablePlugins(SbtWeb)
```
mvexel: @Crashfreak I am not sure if this is necessary - I had persistent problems with client side libraries (lessjs and jshint I believe) not being present, and including SbtWeb here resolved it. Any comments on this?

conf/application.conf
```diff
@@ -34,9 +34,7 @@ application.langs="en"
 # By convention, the default datasource is named `default`
 #
 db.default.driver=org.postgresql.Driver
-db.default.url="jdbc:postgresql://localhost:5432/mp_dev"
-db.default.username=osm
-db.default.password=osm
+db.default.url=${DATABASE_URL}
```
mvexel: @Crashfreak note the removal of the DB defaults from the config file. As noted in the PR comment you now need an envvar like `DATABASE_URL=jdbc:postgresql://osm:osm@localhost:5432/mp_dev`. This is required to make the Heroku build work, and I prefer not to have credentials in the config file anyway.

Crashfreak: This is definitely the correct approach, so good change.

build.sbt
Crashfreak: On my local build I haven't had any issues with it, so probably something I should look into a little more. What problem exactly did the plugin resolve?

PR #22, opened 2016-04-20
Improvements to readme and add automatic down migrations

mvexel:
This PR:

* adds some detail and fixes some small errors in the readme.
* adds automatic applying of downward migrations in the DB to ensure a smooth CI flow


Crashfreak:
Nice.

PR #29, opened 2016-04-28
Fixed search functionality, recent backend change stopped it working,…

Crashfreak:
… and hotkeys were not allowing you to type in the search box.

Added default hashtags for comments in ID and JOSM.
Added some base work for future performance improvements.
Added ability to create challenges and subsequently tasks based on an overpass API query.

### DIFF
--- a/app/controllers/FormEditController.scala
+++ b/app/controllers/FormEditController.scala
@@ -3,24 +3,30 @@ package controllers
 import javax.inject.Inject
 
 import org.maproulette.Config
-import org.maproulette.actions.{ActionManager, Actions}
+import org.maproulette.actions.Actions
 import org.maproulette.controllers.ControllerHelper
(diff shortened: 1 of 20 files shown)


app/controllers/FormEditController.scala
```diff
@@ -87,9 +93,11 @@ class FormEditController @Inject() (val messagesApi: MessagesApi,
         challenge => {
           val id = if (itemId > -1) {
             dalManager.challenge.update(Json.toJson(challenge)(Challenge.challengeWrites), user)(itemId)
+            Future { buildOverpassQLTasks(challenge, user) }
```
mvexel: 👍

Crashfreak: It was actually fairly easy to do and I think works better than I expected. :)

Issue #72, opened 2016-05-25
changing instruction in challenge does not propagate to tasks


mvexel:
When you change the instruction in an existing challenge that has tasks, the task instructions don't get changed as well. 

The assumption should be that unless you specifically override the instruction for a task, it inherits that value from the challenge.

mgcuthbert:
Fixed in recent PR

Issue #68, opened 2016-05-24
Don't reload Overpass query if it didn't change?


mvexel:
If you edit the details of a challenge, the Overpass query will be re-fired upon submitting the changes. This may not be what the user wants. Perhaps make it optional. Or add an 'reload overpass' option to the challenge admin menu to do this explicitly and never do it otherwise. Or add a date interval (every day / week) to schedule re-running the query.

mgcuthbert:
I think a forced reload and schedule for overpass queries would both be good ideas.

mgcuthbert:
Fixed in recent PR, however leaving open because I did not add any scheduling to update the overpass query or the remote url.

mvexel:
Going to close this and create new ticket for scheduling Overpass queries to run at an interval the user can set.

Issue #43, opened 2016-05-12
Import tasks from GeoJSON


mvexel:
It would be neat to have two ways of mass task generation for a challenge: Overpass (implemented) and GeoJSON import. GeoJSON could be either from a Gist ([example](https://gist.github.com/mvexel/a1c8513313d136c83cec89803857499a)), from geojson.io ([example](http://geojson.io/#id=gist:mvexel/975e64532cb5edf2ad28&map=10/53.7724/-2.7775)) or even pasting in a text field?

mvexel:
Or a random URL, example: http://pgh.terrapattern.com/?lat=40.3902895&lng=-80.1144195

mgcuthbert:
Fixed in recent PR.

PR #80, opened 2016-05-29
Updating Challenge creation process

mgcuthbert:
Added a small challenge wizard that will allow users to create challenges based on local geo json (file upload) and remote geo json. (Add a remote url).
Fixed a couple of minor bugs, namely having the task instruction use the challenge instruction if not set in the task. Also the number of tasks that was being displayed in the metrics was incorrect.


mgcuthbert:
This addresses issue #72, issue #68 and issue #43.

PR #141, opened 2016-06-21
Challenge Metrics Update

mgcuthbert:
This PR should improve the challenge metrics page. Fix various bugs related to inconsistent numbers and add a challenge metrics table that gives you the raw details for each challenge and allow you to dive in to each one directly from the metrics page.


mgcuthbert:
Need to close PR in order to update a couple of things in the 3.sql file. I don't like adding this file too often and I have something to add in there.

PR #173, opened 2016-07-14
Removed references to Survey in UI until Survey functionality flushed…

mgcuthbert:
… out a little bit more. It is still accessible in the UI.


mgcuthbert:
The above title is meant to say that it is still accessible in the API, not in the UI.

PR #184, opened 2016-07-22
Changed name of distribution package

mgcuthbert:
There was an issue with deployment that was caused when the MapRouletteV2 distribution package's name was changed due to a version change. I updated the build file to always create the same distribution package name for consistency.

Also added an API call to update all the task priorities.


mvexel:
Oh cool, I was wondering what the deal was with that non-existent ZIP file when I tried to deploy.

Issue #194, opened 2016-08-09
max / default zoom levels


martijnv-telenav:
It appears that the max zoom level set for a challenge is used as the default zoom level and the default zoom level is ignored. 

Also zoom level 19 does not make sense for all base maps, I notice that osm.org tiles do not load consistently on that zoom level. So suggesting that the default zoom level tops out at 18.

ianmcorvidae:
I think this is related somehow to #186 but I'm not completely sure how. Just to be sure whoever fixes either bug is aware of the other!

mgcuthbert:
Resolved in PR #203

Issue #186, opened 2016-07-29
Imagery not showing up


aawiseman:
When I start working on a project the basemap is just gray, no imagery or OSM or anything. I've tried a few different projects. I'm using Safari.

ianmcorvidae:
I've experienced this as well with zoom levels 17 or greater. Seems to work fine on z16 though. I notice that I also can't switch imagery at the closer zooms, so it may be that support for those zoom levels is broken somehow.

aawiseman:
@ianmcorvidae that's what I've experienced too. It shows up when I zoom out, but that's tricky for small features like buildings.

mgcuthbert:
Resolved by PR #203

aawiseman:
@mgcuthbert awesome! thanks

PR #203, opened 2016-09-22
Fixed Zoom Level Bugs

davis20:


mgcuthbert:
This will resolve issues #194 and #186

Issue #218, opened 2016-11-01
Link to user name in toast message


mvexel:
It's neat that we have the last modified user now in the task instruction toast message but it would be great if it was a link to that user's OSM page (`https://osm.org/user/USERNAME`) and there would be a direct link to send that user a message (`https://www.openstreetmap.org/message/new/USERNAME`)

mgcuthbert:
Resolved from PR #224

Issue #202, opened 2016-09-19
Task details returned as string rather than geoJSON


mvexel:
if you retrieve tasks for a challenge, the location and geometries properties are wrapped as strings rather than just plain JSON objects. Example below.

`````` [
  {
    "id": 705625,
    "name": "1",
    "parent": 532,
    "instruction": "",
    "location": "{\"type\":\"Point\",\"coordinates\":[-113.6606344,37.1338572]}",
    "geometries": "{\"type\":\"FeatureCollection\",\"features\":[{\"type\":\"Feature\",\"geometry\":{\"type\":\"Point\",\"coordinates\":[-113.6606344,37.1338572]},\"properties\":{\"@id\": \"node/1438930535\", \"@uid\": \"436419\", \"name\": \"Jacob hamblin home\", \"@user\": \"wvdp\", \"tourism\": \"museum\", \"@version\": \"3\", \"wikipedia\": \"en:Jacob Hamblin House\", \"@changeset\": \"34123009\", \"@timestamp\": \"2015-09-19T15:16:51Z\"}}]}",
    "status": 0,
    "priority": 0
  },
  {
    "id": 705626,
    "name": "2",
    "parent": 532,
    "instruction": "",
    "location": "{\"type\":\"Point\",\"coordinates\":[-113.5815229,37.1101603]}",
    "geometries": "{\"type\":\"FeatureCollection\",\"features\":[{\"type\":\"Feature\",\"geometry\":{\"type\":\"Point\",\"coordinates\":[-113.5815229,37.1101603]},\"properties\":{\"@id\": \"node/4309748792\", \"@uid\": \"245874\", \"name\": \"Daughters of Utah Pioneers Museum\", \"@user\": \"randoogle\", \"tourism\": \"museum\", \"@version\": \"1\", \"@changeset\": \"40831056\", \"@timestamp\": \"2016-07-18T19:01:42Z\"}}]}",
    "status": 0,
    "priority": 0
  }
]```
``````

mgcuthbert:
Resolved in PR #224

Issue #160, opened 2016-07-08
Get in touch with challenge author


mvexel:
It would be super nice to be able to get in touch with the author of a challenge. This would require discussion: an in-app messaging system? just have challenge authors add their email address and provide a link to that? just provide the OSM ID and a link to message that user through OSM?

mgcuthbert:
I wouldn't ordinarily say this, but in this case it would seem like using OSM to pass the message through might be the best option. I am not sure building a message app within MapRoulette would be a beneficial use of time when there is a perfectly acceptable solution in OSM. So we could have something like "Send Message to [OSM Username of Challenge Owner]" and then that link just takes them to the message page in openstreetmap.org. That is my vote anyway.

mvexel:
Makes sense to me! Let's do it that way.

mvexel:
Makes sense. This would relate to #218 .

mgcuthbert:
This is partly resolved in PR #224, I will leave open for now as there may be more area's in the UI that could benefit from including links to the challenge owner.

PR #224, opened 2016-11-15
patch from mgcuthbert

mvexel:
@mgcuthbert please supply patch notes and merge :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ dist
 /.settings
 /.sbtserver
 conf/dev.conf
+project/.sbtserver
+project/.sbtserver.lock
+project/play-fork-run.sbt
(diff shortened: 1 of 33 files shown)


mgcuthbert:
Patch includes the following:
1. Moves zoom control to the right top corner of the map.
2. Modifies instruction to include links to message use who last modified the task.
3. Modifies challenge blurb to include owner and ability to message the owner.
4. Allows ending "/" in url and will redirect to url without "/" with a MovedPermanently status.
5. Added owner id in the challenge, so it is always associated with some user.
6. Updated challenge API creation to allow to create challenges and tasks using a geo json in the API.

mgcuthbert:
Fixes #218, #160 and #202

Issue #93, opened 2016-06-09
'Blurb' not easy to understand


mvexel:
The word `Blurb` does not make a lot of sense and it's not clear how it's used. Let's find a better term and describe how / where it's used.

mgcuthbert:
So interestingly enough this is not used at all in the current UI, however the idea of it would be to describe the Challenge that you are working on, or wanting to work on. However as I said this, this could easily be taken by the description field.

mvexel:
I don't see a great reason to keep it then...Shall we remove it altogether?

In MapRoulette 1 we used it to have a short description (one liner) about the challenge. If we envisage having that in the future we can keep it, in that case I suggest we rename it 'Short Description'.

mgcuthbert:
I think right now we could just remove it, and if we find a requirement later on we can add the new 'Short Description' column in for that need. However I don't currently see a current or future need for it.

mvexel:
Labeling prio 1 because I get questions about the blurb more often than I care for :)

reichg:
Looks like this was resolved in #247

PR #228, opened 2016-11-22
Remove blurb from Challenge wizard #93

andreic-telenav:
I removed blurb from the JS and html in the challenge wizard

related to https://github.com/maproulette/maproulette2/issues/93


mgcuthbert:
I think I incorrectly stated in issue #93 that the blurb was not being used, however as shown in your commit it was being used in at least one place. So currently if this was merged you would not see the blurb in the map challenge view. So we might want to rethink this one a bit.

PR #285, opened 2017-03-04
Enable hotkeys and add cheat sheet

nrotstan:
- Enable hotkeys.
- Add cheat sheet popover that displays available keyboard shortcuts.

mr2_hotkeys


mvexel:
@nrotstan This PR seems to have rendered the search box non-functional -- at least I cannot think of another reason why it stopped working. Could you look into that if possible?

nrotstan:
Yes, will do!

PR #286, opened 2017-03-04
German translation

nebulon42:
I have added a German translation and have corrected some typos in the English messages.


mvexel:
Thanks! Danke!

mvexel:
I added one string `footer.getintouch=Get in touch`

Could you provide a translation for that please?
Thanks,

nebulon42:
Sure. Another question: I saw that you moved away from the colloquial "someone beat me to it" to the more neutral "Already fixed". Was the intention here to formalize or to move away from a phrase that might be harder to understand for non-native speakers?

mvexel:
A bit of both. The `Someone beat me to it` was carried over from MapRoulette 1. Some expressed confusion about what it meant and what pressing that particular button "does" to the task. I think the more neutral and to-the-point `Already fixed` works better. Would you agree?

nebulon42:
Yes, I think that would be better although I liked a bit of informal touch for a gamified software. I will change that in German to a more formal tone too.

PR #325, opened 2017-07-10
Add missed comma in settings file. Fixes #324

### LINKED ISSUE #324: Missed comma in settings results in compilation error
It looks like a comma was missed in settings.scala.html when the FR translation was added, resulting in a compilation error when an attempt is made to load the app. I will submit a pull request.

compilation_error

nrotstan:
Fixes compilation error resulting from missed comma in settings
file when the fr translation was added.


mvexel:
Thanks @nrotstan -- that was an oversight on my part. When I get a chance I will deploy the new translations to production as well.

PR #401, opened 2018-04-14
Add user leaderboard and top-challenges APIs

nrotstan:
User leaderboard API retrieves the top scoring users based on task
completion over a specified timeframe. The number of points awarded for
each completed task varies depending on the completion status. Each
user's top challenges, based on activity, are also included in the
response.

User top-challenges API retrieves the top challenges, based on user's
activity, over a specified timeframe.

### DIFF
--- a/app/org/maproulette/data/DataManager.scala
+++ b/app/org/maproulette/data/DataManager.scala
@@ -58,6 +58,9 @@ case class ChallengeActivity(date:DateTime, status:Int, statusName:String, count
 case class RawActivity(date:DateTime, osmUserId:Long, osmUsername:String, projectId:Long,
                        projectName:String, challengeId:Long, challengeName:String,
                        taskId:Long, oldStatus:Int, status:Int)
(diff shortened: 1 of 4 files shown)


app/org/maproulette/data/DataManager.scala
```diff
@@ -442,6 +445,83 @@ class DataManager @Inject()(config: Config, db:Database)(implicit application:Ap
...
+        userId <- long("users.id")
+        name <- str("users.name")
+        avatarURL <- str("users.avatar_url")
+        score <- int("score")
+      } yield LeaderboardUser(userId, name, avatarURL, score,
+                              this.getUserTopChallenges(userId, start, end))
+
+      SQL"""SELECT users.id, users.name, users.avatar_url, SUM(
+              CASE sa.status
+                WHEN ${Task.STATUS_FIXED} THEN 5          /* points */
```
mgcuthbert: These values should be configurable. And then we would just retrieve what value you want from config (if none found then assume a default).

PR #413, opened 2018-04-25
Support challenge difficulty in extendedFind API.

nrotstan:
Add support for `cd` query parameter to challenges/extendedFind API for
limiting challenge results by difficulty. Defaults to any difficulty,
preserving existing search behavior in absence of new parameter.

### DIFF
--- a/app/org/maproulette/models/dal/ChallengeDAL.scala
+++ b/app/org/maproulette/models/dal/ChallengeDAL.scala
@@ -691,6 +691,11 @@ class ChallengeDAL @Inject() (override val db:Database, taskDAL: TaskDAL,
         case _ =>
       }
 
+      searchParameters.challengeDifficulty match {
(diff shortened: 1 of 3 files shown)


app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -691,6 +691,11 @@ class ChallengeDAL @Inject() (override val db:Database, taskDAL: TaskDAL,
         case _ =>
       }
 
+      searchParameters.challengeDifficulty match {
+        case Some(v) if v != -1 => this.appendInWhereClause(whereClause, s"c.difficulty = ${v}")
```
mgcuthbert: I would say if v > 0 & v < 4, as those are the only valid values for challenge difficulty. (Ie. 1, 2, 3)

PR #436, opened 2018-06-18
project-manager administration API enhancements.

nrotstan:
Additions and modifications to APIs to help support administration of
project managers:

* Add GET `user/project/:projectId` API for retrieving the managers of a
project.
* Add GET `users/find/:username` API for looking up users by OSM
username. Results are limited to just a few public OSM profile fields.
* Change PUT `/user/project/{projectId}/{groupType}` -- which **adds** a
group type -- to use POST instead.
* Add new PUT `/user/project/{projectId}/{groupType}` that **replaces**
any existing group type with the given group type.
* Modify DELETE `/user/project/{projectId}/{groupType}` to support -1
value for groupType to indicate that all group types should be removed
from the user.

### DIFF
--- a/app/org/maproulette/session/dal/UserDAL.scala
+++ b/app/org/maproulette/session/dal/UserDAL.scala
@@ -102,6 +102,27 @@ class UserDAL @Inject() (override val db:Database,
     }
   }
 
+  // The anorm row parser to convert user records to search results
+  val searchResultParser: RowParser[UserSearchResult] = {
+    get[Long]("users.osm_id") ~
+    get[String]("users.name") ~
+    get[Option[String]]("users.avatar_url") map {
+      case osmId ~ displayName ~ avatarURL =>
+        UserSearchResult(osmId, displayName, avatarURL.getOrElse("")),
+    }
+  }
+  // The anorm row parser to convert user records to project manager
+  val projectManagerParser: RowParser[ProjectManager] = {
+    get[Long]("project_id") ~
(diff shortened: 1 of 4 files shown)


app/org/maproulette/session/dal/UserDAL.scala
```diff
@@ -473,26 +513,85 @@ class UserDAL @Inject() (override val db:Database,
...
+    * Sets the group type of a user in a project, first deleting any prior group types,
+    * in a single transaction.
+    *
+    * @param osmID The OSM ID of the user
+    * @param projectId The id of the project to remove the user from
+    * @param groupType The type of group to set 1 - Admin, 2 - Write, 3 - Read
+    * @param user The user making the request
+    * @param c
+    */
+  def setUserProjectGroup(osmID:Long, projectId:Long, groupType:Int, user:User)(implicit c:Connection=null) : Unit = {
```
mgcuthbert: I don't fully understand the use case around this function. Why would we need something to delete all user groups and then add them to just one. Is this like a sort of reset function?

mgcuthbert: Ok I think this clears up things for me. For some reason I thought that the SQL statement was deleting the users association in other projects as well. I misread the SQL query. So I think this makes sense, as the permissions model is setup so that if you have Admin access you have write and read access, and if you have write access you have read access. So limiting to a single group type isn't an issue. I think this looks good, and thank you for the detailed explanation.

PR #468, opened 2018-10-19
New STATUS_FINISHED for challenges with no more incomplete tasks

krotstan:
* Rename STATUS_COMPLETE to STATUS_READY to avoid potential confusion

* setTaskStatus sets challenge status to STATUS_FINISHED if no more tasks
remain with CREATED status

* Add evolution that sets existing challenges to STATUS_FINISHED if their tasks
are all complete

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -435,6 +435,8 @@ class TaskDAL @Inject()(override val db: Database,
     }
 
     this.cacheManager.withOptionCaching { () => Some(task.copy(status = Some(status))) }
+    this.challengeDAL.get().updateFinishedStatus()(task.parent)
+
     updatedRows
   }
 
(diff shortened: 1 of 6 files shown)


krotstan:
I accidentally cleaned up the wrong branch and didn't mean to close this PR.

app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -194,13 +194,22 @@ class TaskDAL @Inject()(override val db: Database,
...
-      this.mergeUpdate(cachedItem.copy(name = name,
+      val task = this.mergeUpdate(cachedItem.copy(name = name,
         parent = parentId,
         instruction = Some(instruction),
         status = Some(status),
         geometries = geometries,
         priority = priority,
         changesetId = Some(changesetId)), user)
+
+      if (status == Task.STATUS_CREATED) {
```
mgcuthbert: So it might be a little more efficient to have a scheduled job, update the status for the challenge like once a day or something like that, instead of having this call made to the database on every single task update. Unless you need to have the "finished" status immediately.

krotstan: We do need the "finished" status immediately to know whether to show the user a "challenge complete" message.

PR #469, opened 2018-10-26
Changed project find to also search by display_name

krotstan:
* Override find method in ProjectDAL to also search by display_name
  in addition to name.

### DIFF
--- a/app/org/maproulette/models/dal/ProjectDAL.scala
+++ b/app/org/maproulette/models/dal/ProjectDAL.scala
@@ -10,6 +10,7 @@ import anorm.SqlParser._
 import org.joda.time.DateTime
 import org.maproulette.Config
 import org.maproulette.cache.CacheManager
+
 import org.maproulette.exception.UniqueViolationException
 import org.maproulette.models._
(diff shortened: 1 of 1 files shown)


app/org/maproulette/models/dal/ProjectDAL.scala
```diff
@@ -147,6 +148,29 @@ class ProjectDAL @Inject() (override val db:Database,
...
+  /**
+    * This find function will search the "name" and "display_name" fields for
+    * any references of the search string. So will be wrapped by %%, eg. LIKE %test%
+    *
+    * @param searchString The string to search for within the name field
+    * @param limit Limit the number of results to be returned
+    * @param offset For paging, ie. the page number starting at 0
+    * @return A list of tags that contain the supplied prefix
+    */
+  override def find(searchString:String, limit:Int = Config.DEFAULT_LIST_SIZE, offset:Int = 0, onlyEnabled:Boolean=false,
```
mgcuthbert: It seems like you might want to think about using this.fuzzySearch, instead of just the basic search option.

PR #478, opened 2018-12-03
Add support for returning leaderboard with rank for specific user.

krotstan:
* Add new method to return a specific user's leaderboard that
  also includes leaderboards for users that bracket that user's rank


app/org/maproulette/data/DataManager.scala
```diff
@@ -94,6 +94,44 @@ class DataManager @Inject()(config: Config, db:Database)(implicit application:Ap
...
+
+  /**
+   * Returns the SQL to fetch the ordered leaderboard data with rankings. Can be filtered
+   * by a list of projects, challenges, users and start/end date.
+   **/
+  private def leaderboardWithRankSQL(userFilter:Option[List[Long]]=None, projectFilter:Option[List[Long]]=None,
+                         challengeFilter:Option[List[Long]]=None, start:Option[DateTime]=None,
+                         end:Option[DateTime]=None) : String = {
+    s"""
+        SELECT users.id, users.name, users.avatar_url, ${this.scoreSumSQL()} AS score,
```
mgcuthbert: What kind of performance do we get with this query? Considering that the status_actions table can get pretty large.

krotstan: I've added a new commit that includes a job that will generate new tables to improve performance (default cases only -- past month, 3 months, 6 months, year and no project/challenge/user filters). When this is deployed to production the conf file will need to include the interval for which this job should run. The job takes around 2.4 seconds to run on my average machine over 430k rows.

challengesLeaderboard.interval = "20 minutes"

PR #488, opened 2019-01-14
Phase 1 commit for Suggested Fixes.

mgcuthbert:
This commit introduces API for Suggested Fixes for Tags only. It adds two new API's, one to test changes and the other to submit changes. It handles creation and closing of Changesets in the background and users focus entirely on tag delta changes that are submitted to the server.


mgcuthbert:
Yes, I accidentally did, I need to add a gitignore or something to that.

PR #506, opened 2019-02-22
Add support for a review system

krotstan:
* Add two flags to user: needsReview, isReviewer

* When setTaskStatus is called update whether this task
  needs to be reviewed. Add new request review flag to
  override user's profile setting of 'needsReview'

* Return new review fields on task

* Make available API method to get tasks that need review

* Make available API method to get tasks that have been
  reviewed (either as the reviewer or as the mapper)

* Make available API method to set task review status

* Add 'requestClaimedBy' to ensure two reviewers aren't
  reviewing same task at the same time.

* Add job to clean claims older than 12 hours.

* Make available API method to start/cancel review (claim)


app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -221,6 +225,121 @@ class TaskController @Inject()(override val sessionManager: SessionManager,
...
+      }
+    }
+  }
+
+  /**
+   * Fetches the matching parent object and inserts it (id, name, status)
+   * into the JSON data returned. Also fetches and inserts usernames for
+   * 'reviewRequestedBy' and 'reviewBy'
+   */
+  private def _insertExtraJSON(tasks: List[Task]): JsValue = {
```
mgcuthbert: This kind of thing shows me how much we need a graphQL server.

mgcuthbert:
So MapRoulette currently doesn't have a great set of unit tests. However there are a pretty good set of Postman Collection runner "tests" or integration tests that test the API against a running system. With these kinds of changes it would be great if a) you can run those tests to make sure that no harm done kind of thing. And b) add new tests that validate the functionality of the newly added stuff here.

app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -221,6 +225,151 @@ class TaskController @Inject()(override val sessionManager: SessionManager,
...
+    *
+    * @return
+    */
+  def nextTaskReview(sort:String, order:String) : Action[AnyContent] = Action.async { implicit request =>
+    this.sessionManager.authenticatedRequest { implicit user =>
+      SearchParameters.withSearch { implicit params =>
+        //cs => "my challenge name"
+        //o => "mapper's name"
+        //r => "reviewer's name"
+        println("NEXT TASK REVIEW. Params?")
```
mgcuthbert: These look like lines you would insert to debug something. Probably shouldn't check this in.

app/org/maproulette/models/Task.scala
```diff
@@ -40,9 +40,15 @@ case class Task(override val id: Long,
                 geometries: String,
                 suggestedFix: Option[String] = None,
                 status: Option[Int] = None,
-                priority: Int = Challenge.PRIORITY_HIGH,
+                mappedOn: Option[DateTime] = None,
```
mgcuthbert: Should this maybe be stored in it's own table. Like. TaskReview table that just references the Task table and can be retrieved when required?

conf/evolutions/default/28.sql
```diff
@@ -0,0 +1,141 @@
+# --- MapRoulette Scheme
+
+# --- !Ups
+
+-- Add Mapped on timestamp
+ALTER TABLE "tasks" ADD COLUMN mapped_on timestamp without time zone DEFAULT NULL;;
```
mgcuthbert: See comments above. I think we should consider a separate table for task reviews.

PR #523, opened 2019-03-26
Change migration sql on needsReview default bug

krotstan:
Migration sql was not correctly converting the needsReview
flag from boolean to integer as expected


mgcuthbert:
Why exactly do we need to convert from boolean to int?

krotstan:
So that we can have a way to "lock" the setting so a user cannot opt out of being a reviewer if the admin sets it.

PR #536, opened 2019-04-18
Fix 'modify task' sql error by casting correct type

krotstan:
When trying to 'modify task' and the SQL function 'create_update_task' was being called, it was throwing a sql error as two of the types were mismatched. Add cast to make these types match as expected. (Fixes #731)


danielduhh:
Hi @krotstan! Just curious: was there a database update to go with this as well?

krotstan:
No database update needed.

PR #540, opened 2019-04-30
Add API to start/release (lock) a task

krotstan:
Add ability to lock and unlock a task via new methods startOnTask and releaseTask.


app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -547,6 +547,49 @@ class TaskDAL @Inject()(override val db: Database,
     }
   }
 
+  /**
```
mgcuthbert: I think this work would better in the TaskController. Instead of calling this function, I think we should call the lockItem and unlockItem directly from the API calls startOnTask and releaseTask.

PR #542, opened 2019-05-07
Fix review requested tasks query 

krotstan:
After adding project filter, the inner join was joining on projects twice in some cases.


mgcuthbert:
Are there any cases where it was not joining twice?

krotstan:
If you were a super user is worked fine. The second join on projects was happening for non-super users in the "count" query (the join on project was already included in ${joinClause}).

PR #543, opened 2019-05-07
Add user metrics support

krotstan:
* Make leaderboard able to fetch even if not logged in

* Add API getUserMetrics

* Change getUser (and by osm username) to allow a user not
  logged in (also not super user) to fetch some basic user
  information like name, avatar, etc.

### DIFF
--- a/app/org/maproulette/controllers/api/UserController.scala
+++ b/app/org/maproulette/controllers/api/UserController.scala
@@ -60,71 +60,57 @@ class UserController @Inject()(userDAL: UserDAL,
   }
 
   def getUser(userId: Long): Action[AnyContent] = Action.async { implicit request =>
-    this.sessionManager.userAwareRequest { implicit user =>
-      var result : Option[Result] = None
-
-      user match {
-        case Some(requestUser) =>
-          // User can get itself
-          if (userId == requestUser.id || userId == requestUser.osmProfile.id) {
-            result = Option(Ok(Json.toJson(User.withDecryptedAPIKey(requestUser)(crypto))))
+    this.sessionManager.authenticatedRequest { implicit user =>
+      if (userId == user.id || userId == user.osmProfile.id) {
+        Ok(Json.toJson(User.withDecryptedAPIKey(user)(crypto)))
+      } else if (user.isSuperUser) {
+        this.userDAL.retrieveByOSMID(userId, user) match {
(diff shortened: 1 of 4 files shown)


app/org/maproulette/controllers/api/UserController.scala
```diff
@@ -60,34 +60,71 @@ class UserController @Inject()(userDAL: UserDAL,
...
-    this.sessionManager.authenticatedRequest { implicit user =>
-      if (userId == user.id || userId == user.osmProfile.id) {
-        Ok(Json.toJson(User.withDecryptedAPIKey(user)(crypto)))
-      } else if (user.isSuperUser) {
-        this.userDAL.retrieveByOSMID(userId, user) match {
-          case Some(u) => Ok(Json.toJson(User.withDecryptedAPIKey(u)(crypto)))
-          case None => this.userDAL.retrieveById(userId) match {
-            case Some(u) => Ok(Json.toJson(User.withDecryptedAPIKey(u)(crypto)))
-            case None => throw new NotFoundException(s"No user found with id '$userId'")
+    this.sessionManager.userAwareRequest { implicit user =>
```
mgcuthbert: So the one problem I have with this is that if there is any kind of bug in this request a user's API key could be exposed by calling this method. I think what I would prefer to see is a separate API call to get basic user information. So for example another function call named "getBasicUser". This way even if this particular PR has no bug we protect ourselves from future bugs that could have bad security concerns.

app/org/maproulette/controllers/api/UserController.scala
```diff
@@ -60,34 +60,71 @@ class UserController @Inject()(userDAL: UserDAL,
...
-      if (user.name == username) {
-        Ok(Json.toJson(user))
-      } else {
-        // we don't need to check access here as the API only allows super users to make the call,
-        // so if not a super user, the correct IllegalAccessException will be thrown
-        this.userDAL.retrieveByOSMUsername(username, user) match {
-          case Some(u) => Ok(Json.toJson(u))
-          case None => throw new NotFoundException(s"No user found with OSM username '$username'")
-        }
+    this.sessionManager.userAwareRequest { implicit user =>
```
mgcuthbert: Same comment here as the one above, I think we need to separate out retrieval of basic user information.

PR #559, opened 2019-06-06
Delete tasks asynchronously and in batches

nrotstan:
> Note: while making these changes, I discovered that there was a missing parenthesee that caused bad SQL to be generated if a status filter was given, so I suspect that this API endpoint isn't getting much (if any) use at the moment. Also, given the current webserver configuration, any attempt to delete more than a trivial number of tasks would likely result in a gateway timeout.

* Add new DELETING_TASKS challenge status

* Change the `DELETE /challenge/:id/tasks` API endpoint to delete tasks
asynchronously and to change the challenge status to DELETING_TASKS
while the delete operation is ongoing

* Change `ChallengeDal.deleteTasks` to work in small batches

* Add guard conditions to ChallengeController to prevent bulk task
deletion or challenge rebuilds while a deletion or rebuild is already in
progress

### DIFF
--- a/app/org/maproulette/controllers/api/ChallengeController.scala
+++ b/app/org/maproulette/controllers/api/ChallengeController.scala
@@ -348,8 +348,30 @@ class ChallengeController @Inject()(override val childController: TaskController
     */
   def deleteTasks(challengeId: Long, statusFilters: String = ""): Action[AnyContent] = Action.async { implicit request =>
     this.sessionManager.authenticatedRequest { implicit user =>
-      this.dal.deleteTasks(user, challengeId, Utils.split(statusFilters).map(_.toInt))
-      Ok
+      dalManager.challenge.retrieveById(challengeId) match {
+        case Some(c) =>
+          permission.hasWriteAccess(ProjectType(), user)(c.general.parent)
+          if (c.status.get == Challenge.STATUS_DELETING_TASKS) {
+            throw new InvalidException("Task deletion already in-progress for this challenge")
+          }
+          else if (c.status.get == Challenge.STATUS_BUILDING) {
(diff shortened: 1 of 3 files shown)


mgcuthbert:
I am assuming that this is being used because you actually don't want to delete the challenge itself, as the database as a cascade delete on challenges that will remove the tasks. So if you just want to get rid of the challenge you can just delete the challenge directly. Just want to make sure on that one.

app/org/maproulette/controllers/api/ChallengeController.scala
```diff
@@ -348,7 +348,22 @@ class ChallengeController @Inject()(override val childController: TaskController
...
     this.sessionManager.authenticatedRequest { implicit user =>
-      this.dal.deleteTasks(user, challengeId, Utils.split(statusFilters).map(_.toInt))
+      val originalStatus = dalManager.challenge.retrieveById(challengeId) match {
+        case Some(c) => c.status
+        case None => throw new NotFoundException(s"No challenge found with id $challengeId")
+      }
+
+      // Deleting a lot of tasks can be time consuming, so perform this asynchronously
+      // Todo: perhaps introduce a DELETING status rather than co-opting BUILDING
+      dalManager.challenge.update(Json.obj("status" -> Challenge.STATUS_BUILDING), user)(challengeId)
```
mgcuthbert: I very much agree with this todo. Any reason why you don't just add that status in this PR?

nrotstan: Will do ;-)

nrotstan:
This was so that challenge tasks in specific statuses could be deleted.

nrotstan:
@mgcuthbert New DELETING_TASKS challenge status added.

PR #567, opened 2019-06-19
Fix task retrieval by name

nrotstan:
* Override retrieveByName in TaskDAL to ensure that the task_review
table is joined in

* This fixes the 500 error encountered when using the GET
`challenge/:id/task/:name` API endpoint

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -181,6 +181,27 @@ class TaskDAL @Inject()(override val db: Database,
   }
 
   /**
+    * Retrieves the object based on the name, this function is somewhat weak as there could be
+    * multiple objects with the same name. The database only restricts the same name in combination
+    * with a parent. So this will just return the first one it finds. With caching, so if it finds
+    * the object in the cache it will return that object without checking the database, otherwise
+    * will hit the database directly.
+    *
+    * @param name The name you are looking up by
+    * @return The object that you are looking up, None if not found
+    */
+  override def retrieveByName(implicit name: String, parentId: Long = (-1), c: Option[Connection] = None): Option[Task] = {
+    this.cacheManager.withOptionCaching { () =>
+      this.withMRConnection { implicit c =>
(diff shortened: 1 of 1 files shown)


app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -180,6 +180,27 @@ class TaskDAL @Inject()(override val db: Database,
...
+    * the object in the cache it will return that object without checking the database, otherwise
+    * will hit the database directly.
+    *
+    * @param name The name you are looking up by
+    * @return The object that you are looking up, None if not found
+    */
+  override def retrieveByName(implicit name: String, parentId: Long = (-1), c: Option[Connection] = None): Option[Task] = {
+    this.cacheManager.withOptionCaching { () =>
+      this.withMRConnection { implicit c =>
+        val query = s"SELECT $retrieveColumnsWithReview FROM ${this.tableName} " +
```
mgcuthbert: Are we always returning "$retrieveColumnsWithReivew" and if so should we maybe not just switch it to "$retrieveColumns" and be done with it?

nrotstan: The problem is that we have to join in those review columns from the `task_review` table in the queries, which of course none of the BaseDAL methods do, so we'd have to override every method in BaseDAL (possessing a query) to make it work as $retrieveColumns. Which we can do.

However, I think a strong argument could be made for folding those `task_review` columns into the `tasks` table as they are essentially additional review-related task fields. That would eliminate the joins and the need for the $retrieveColumnsWithReview, and essentially allow everything to just work naturally. There are currently six columns on `task_review`, not counting the id and task_id foreign key:

```
 review_status       | integer
 review_requested_by | integer
 reviewed_by         | integer
 reviewed_at         | timestamp without time zone
 review_claimed_at   | timestamp without time zone
 review_claimed_by   | integer
```

I do agree that we should commit to one way or the other, as otherwise we're just going to run into additional bugs like this one down the road when some execution path makes use of a BaseDAL method that doesn't pull in those columns that are expected by something downstream.

PR #575, opened 2019-06-21
Change random tasks to not return ones locked by user.

krotstan:
If a task is locked by a user, it means the user was currently
(or recently) working on it. Change getRandomTasks to exclude
these so that we are always fetching brand new tasks the user
hasn't seen before.


mgcuthbert:
What is exactly the use case that required this change?

krotstan:
When asking for the next task to work on, we don't want the same task back that was just being worked on (which is the locked task). Also, the updating of the status of the task is done as a future, so theoretically there is a race condition to make sure the task status is updated to fix before you ask for the next task to work on.

krotstan:
Closing this pull request and opening a new one since this branch is so behind.

PR #577, opened 2019-06-27
Update modified date in cache in setTaskStatus

krotstan:
Modified date does get updated during setTaskStatus but the new timestamp isn't updated in the cache so a subsequent call to get the task details does not correctly show an updated modified time.

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -645,10 +645,12 @@ class TaskDAL @Inject()(override val db: Database,
     if (reviewNeeded) {
       this.cacheManager.withOptionCaching { () => Some(task.copy(status = Some(status),
                                                  reviewStatus = Some(Task.REVIEW_STATUS_REQUESTED),
-                                                 reviewRequestedBy = Some(user.id))) }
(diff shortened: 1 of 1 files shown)


app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -645,10 +645,12 @@ class TaskDAL @Inject()(override val db: Database,
     if (reviewNeeded) {
       this.cacheManager.withOptionCaching { () => Some(task.copy(status = Some(status),
                                                  reviewStatus = Some(Task.REVIEW_STATUS_REQUESTED),
-                                                 reviewRequestedBy = Some(user.id))) }
+                                                 reviewRequestedBy = Some(user.id),
+                                                 modified = new DateTime())) }
```
mgcuthbert: This should be handled by a trigger in the database. So if it isn't working we need to check there.

krotstan: The database is correctly updating the modified timestamp. The problem was that a the cache object being returned did not have the changed field -- so this change only updates the modified date on the cache object so that what's immediately returned makes sense. Next time that cache object is refreshed it will get the value from the database.

PR #580, opened 2019-07-17
Changed geojson export to list maproulette props as mr-*

krotstan:
* Moved geojson export properties from under 'maproulette' to top
  level properties mr-* so that they work in JOSM

* Added translation from task status int to text

### DIFF
--- a/app/org/maproulette/models/dal/ChallengeDAL.scala
+++ b/app/org/maproulette/models/dal/ChallengeDAL.scala
@@ -729,25 +729,34 @@ class ChallengeDAL @Inject()(override val db: Database, taskDAL: TaskDAL,
                    FROM ( SELECT 'Feature' As type,
                                   ST_AsGeoJSON(lg.geom)::json As geometry,
                                   hstore_to_json(lg.properties)::jsonb ||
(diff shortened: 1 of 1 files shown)


app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -729,25 +729,34 @@ class ChallengeDAL @Inject()(override val db: Database, taskDAL: TaskDAL,
                    FROM ( SELECT 'Feature' As type,
                                   ST_AsGeoJSON(lg.geom)::json As geometry,
                                   hstore_to_json(lg.properties)::jsonb ||
-                                  json_build_object('maproulette',
+
```
mgcuthbert: This may just be a matter of me not quite seeing what is going on here. But where did the json_build_object function go to exactly? Or do we just not need it anymore? I can't remember the specifics of this sql so apologize if this is a stupid question.

krotstan:
The json build object was to construct a new json object under “maproulette” like:

Maproulette:{ taskStatus: fixed, etc}

But it was asked that the fields get moved up as top level properties.

PR #593, opened 2019-08-08
Add flag for challenge URL inclusion.

zerebubuth:
This adds a flag to the Challenge type and database. When true, the URL of the challenge can be appended to the changeset comment, which is useful for easier traceability of changesets back to the challenge intent and instructions.

This is optional, defaulted to false for the moment. There are corresponding changes to the React UI which allow it to be toggled at challenge creation time. In contrast to the hashtag, which is added before the challenge is created, the URL needs to know the challenge ID and so must be added later. Rather than have the UI read back the new ID and re-save the challenge, it seemed cleaner to store the flag.

Refs https://github.com/osmlab/maproulette3/issues/178


mgcuthbert:
This PR has been opened a while and requires branches to be merged.

mgcuthbert:
I am closing this PR, as unfortunately it is pretty old and out of date.

PR #604, opened 2019-08-21
Suggested fixes mr3 integration

krotstan:
Allow support for challenges with Suggested Fixes  …
* Suggested Fixes can be included in either the task JSON as
another field "suggestedFix" or can be included in the
geometry geoJSOn properties under a "maproulette" field
called "suggestedFix".

Challenge will include a hasSuggestedFix flag to indicate
this challenge has suggeted fixes.

Sample suggestedFix:
suggestedFix: {
	osmId: "243499890" // not needed if in geoJSON
	osmType: "way"     // not needed if in geoJSON
	update: {
		tags: {
			add: [{name:"restaurant", value:"yes"}],
			update: [{name:"building", value:"yes"}]
			delete: [{name:"bldg", ifUnused:"true"}]
		}
	}
}

Tasks returned will included a suggestedFix property.

* To apply suggested Fix to a task use the api applyTagFix.
This will:
  - Submits tag change to OSM
  - Updates task status to FIXED

* During testing if you do not want the changes to be sent
to OSM (or dev OSM) then add 'mockOSM' config option.

osm {
  mockOSM=true
}


app/org/maproulette/Config.scala
```diff
@@ -37,6 +37,8 @@ class Config @Inject()(implicit val configuration: Configuration) {
     this.config.getOptional[Boolean](Config.KEY_DEBUG).getOrElse(false)
   lazy val isDevMode: Boolean =
     this.config.getOptional[Boolean](Config.KEY_DEVMODE).getOrElse(false)
+  lazy val mockOSM: Boolean =
```
mgcuthbert: I think you should change the variable name here to reference the fact that you are using this to mock changeset submission to OSM. Otherwise it might be confusing about what this actually does.

PR #607, opened 2019-09-03
Add flag to not return too hard tasks

krotstan:
- Add configuration flag skipTooHard to not return too hard
  tasks when asking for the next random task.


matkoniecz:
Any idea where I can find it in the UI? I wanted to propose this functionality (for everything) but I would want to at least enable it for my tasks.

nrotstan:
If you're running your own server, you'd set it in the backend .conf file, @matkoniecz. It applies to all tasks server-wide. We don't have it enabled on maproulette.org, however. There's more discussion about treatment of too-hard tasks in osmlab/maproulette3#217

mvexel:
I'm reviving the linked MR3 ticket and flagging it for implementation in the front end in the foreseeable future. Just need to scope it with @jschwarz2030 and see if we want to include it in a next point release or the next major release.

PR #616, opened 2019-09-23
Remove hour wait for tasks in created status when fetching next

krotstan:
When fetching next tasks to work on, we only return tasks that
the user has not worked on in the last hour. If the task's
status is reset back to created this limit still applies. Do
not require an hour wait limit if the task is in a created
status so it can be worked on right away in this instance.


mgcuthbert:
Removing this can be problematic, I mean I get the reason, however you might instead of removing the limit, rather include a check on the modified field and if the task has been recently modified then remove the limit.

PR #617, opened 2019-09-23
On bulk status update remove tasks from affected bundles

krotstan:
If doing a bulk update of task statuses then remove from
any bundles that the tasks are a part of and also other
tasks in the bundle.

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -312,6 +312,19 @@ class TaskDAL @Inject()(override val db: Database,
         this.challengeDAL.get().updateFinishedStatus()(parentId)
       }
 
+      task match {
+        case Some(t) =>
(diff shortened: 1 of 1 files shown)


app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -344,6 +344,12 @@ class TaskDAL @Inject()(override val db: Database,
       val result = extractSuggestedFix(element.parent, element.geometries, element.suggestedFix)
       val geometries = result._1
       var suggestedFix = result._2
+
+      if (element.bundleId != None) {
+        SQL(s"UPDATE tasks SET bundle_id = NULL, is_bundle_primary = NULL " +
```
mgcuthbert: I feel like I am reading this one wrong. Why would you set the bundle_id to NULL if the element task that is supplied has a value for the bundleId?

PR #621, opened 2019-09-27
Fixing initial setup

mvexel:
* Restructured README completely
* fixed evolution 40 which gave errors with `CREATE OR REPLACE`
* Created `dev.conf.example` as a template to create user's own `dev.conf` from
* deleted `dev.conf` from repo


conf/evolutions/default/40.sql
```diff
@@ -4,7 +4,8 @@
 
 -- Creates or updates and task. Will also check if task status needs to be updated
 -- This change adds the mapped_on, review_status, review_requested_by, reviewed_by
-CREATE OR REPLACE FUNCTION create_update_task(task_name text,
+DROP FUNCTION create_update_task(text,bigint,text,integer,jsonb,jsonb,bigint,integer,bigint,text,timestamp without time zone,integer,integer,integer,timestamp without time zone);
```
mgcuthbert: I would suggest using `DROP FUNCTION IF EXISTS ...` just in case.

PR #622, opened 2019-10-01
Add support for a challenge data origin date.

krotstan:
* Add new column on challenge data_origin_date

* When creating a challenge, this field can be include to indicate when the data was sourced

* LastTaskRefresh is now always updated when tasks are rebuilt


mgcuthbert:
So the data origin date, is not something that is automatically retrieved, it is inputted by the user? Should we maybe update the ChallengeService class to add the current date/time when overpass queries are executed?

krotstan:
Correct, the origin data date is supplied by the user -- unless it's an overpass query and then when markTasksRefreshed is called the origin data date will get set to the current date/time.

PR #632, opened 2019-10-17
Handle suggested fixes added as feature properties

nrotstan:
* Ensure suggested-fix JSON added as a feature property is converted to
string form (as all feature properties are)

* Handle unescaping/reconversion back to JSON of stringified
suggested-fix data during task creation


nrotstan:
This fixes osmlab/maproulette3#945

PR #649, opened 2019-11-21
Add API for refreshing task locks

nrotstan:
* Add `task/:id/refreshLock` API endpoint for refreshing an active
task lock, thereby extending its allowed duration

* Add `refreshItemLock` method to Locking trait that updates the lock's
`locked_time` to the current time

* Add evolution that adds `created` column to `locked` table for
tracking the creation time of the lock, allowing the total lock duration
to be tracked even if a lock is refreshed multiple times

* Add basic Postman tests for starting a task, refreshing the task lock,
and releasing the task lock


mgcuthbert:
So while you are in this part of the code, could you also make the lock time configurable. Currently it is 1hr, and we should continue to default to that, but making it default to different times could be useful. `SchedulerActor.scala line 76.

nrotstan:
Good idea, @mgcuthbert , support added.

mgcuthbert:
Thanks.

PR #658, opened 2019-12-12
Add unique constraint to table virtual_project_challenges

krotstan:
* Add unique constraint to virtual_project_challenges so that
  the same challenge can't be added to the same virtual project
  more than once.

* Remove existing duplicate entries.

* In the addChallenge method, don't do the sql insert if the
  challenge has already been added.

### DIFF
--- a/app/org/maproulette/models/dal/VirtualProjectDAL.scala
+++ b/app/org/maproulette/models/dal/VirtualProjectDAL.scala
@@ -55,8 +55,10 @@ class VirtualProjectDAL @Inject()(override val db: Database,
         SQL(query).execute()
       } catch {
           case e:PSQLException if (e.getSQLState == "23505") => //ignore
-          case _ =>
-            throw new InvalidException(s"Unable to add challenge ${challengeId} to Virtual Project ${projectId}.")
+          case other:Throwable =>
+            throw new InvalidException(
+              s"Unable to add challenge ${challengeId} to Virtual Project ${projectId}. " +
+              other.getMessage)
       }
       None
     }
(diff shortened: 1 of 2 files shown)


app/org/maproulette/models/dal/VirtualProjectDAL.scala
```diff
@@ -43,13 +43,27 @@ class VirtualProjectDAL @Inject()(override val db: Database,
...
+           WHERE project_id = {projectId} AND challenge_id = {challengeId} limit 1
+        """)
+          .on('projectId -> projectId, 'challengeId -> challengeId)
+          .as(SqlParser.long("id").singleOpt)
+
+      existing match {
+        case Some(e) => // already in database, do not add again
+        case _ =>
+          val query =
+            s"""INSERT INTO virtual_project_challenges (project_id, challenge_id)
```
mgcuthbert: So you don't have to do the check first. You can instead use a try catch around this insert and then just ignore any unique key violations. That way you only need to make a single call to the database instead of 2 every time. So something like:
```
try {
    ...
} catch {
    case e:PSQLException if (e.getSQLState == "23505") => //ignore
    case _ => // throw some exception here
}
```

app/org/maproulette/models/dal/VirtualProjectDAL.scala
```diff
@@ -46,23 +47,16 @@ class VirtualProjectDAL @Inject()(override val db: Database,
...
-                          VALUES ($projectId, $challengeId)"""
-          SQL(query).execute()
+      try {
+        val query =
+          s"""INSERT INTO virtual_project_challenges (project_id, challenge_id)
+                        VALUES ($projectId, $challengeId)"""
+        SQL(query).execute()
+      } catch {
+          case e:PSQLException if (e.getSQLState == "23505") => //ignore
+          case _ =>
```
mgcuthbert: You should supply the actual exception as well and throw it with the message, otherwise you are essentially ignoring the original issue. This was my bad, as my previous example showed it done exactly this way, but looking at it now, I think it would be bad if we just swallowed the exception.

PR #664, opened 2020-01-07
Parse center property of the overpass output

Zverik:
Should fix https://github.com/osmlab/maproulette3/issues/954: when using `out center` in an overpass query, it would pick up these centers.

Untested locally, since I don't have scala and I have no idea how to get the entire thing running. Sorry for that, fell free to close the PR if that's a burden to the maintainers.


mgcuthbert:
Awesome, thanks for the fix. I just need to test it locally to make sure it works, but it looks good.

mgcuthbert:
I have tested this locally, and looks great. Thanks @Zverik for the fix.

PR #670, opened 2020-01-15
Add support for including progress by priority.

krotstan:
* Add new flag to getProjectSummary called getChallengeSummary
  'includeByPriority' which will return a field 'priorityActions'
  that maps priority to the summarized actions just for that priority.

Example return:
  [{"name":"Winthrop Challenge",
      "id":4070,
       "actions": 
          {"total":110,
            "available":82,
            "fixed":8, etc.},
       "priorityActions": 
        {"2":
            {"total":0,
             "available":0,
              "fixed":0, etc.},
         "1":
            {"total":108,
             "available":81,
             "fixed":8, etc.},
         "0":
            {"total":2,
             "available":1, etc.}
        }
}]

### DIFF
--- a/app/org/maproulette/controllers/api/DataController.scala
+++ b/app/org/maproulette/controllers/api/DataController.scala
@@ -176,6 +176,28 @@ class DataController @Inject()(sessionManager: SessionManager, challengeDAL: Cha
     }
   }
 
+  private def _fetchPrioritySummaries(challengeId: Option[Long], params: Option[SearchParameters],
+                                      onlyEnabled: Boolean = false): mutable.Map[String, JsValue] = {
+    val prioritiesToFetch = List(Challenge.PRIORITY_HIGH, Challenge.PRIORITY_MEDIUM, Challenge.PRIORITY_LOW)
+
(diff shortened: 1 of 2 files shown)


app/org/maproulette/controllers/api/DataController.scala
```diff
@@ -183,19 +185,67 @@ class DataController @Inject()(sessionManager: SessionManager, challengeDAL: Cha
         ))
       } else {
         SearchParameters.withSearch { implicit params =>
-          Ok(Json.toJson(
-            this.dataManager.getChallengeSummary(challengeId = Some(id), priority = Utils.toIntList(priority), params = Some(params))
-          ))
+          val response = this.dataManager.getChallengeSummary(challengeId = Some(id), priority = Utils.toIntList(priority), params = Some(params))
+
+          if (includeByPriority) {
```
mgcuthbert: This block of code is duplicated, so you should move it into a function with some input parameters to handle the input parameters. The Utils.insertIntoJson section would obviously not be included in that.

krotstan: You are right. Extracted it into a separate method.

PR #679, opened 2020-01-24
Add isOSMUserId param to project manager APIs

nrotstan:
> Note: this represents a breaking change to the API that will require
clients who currently pass OSM user ids to the project manager APIs to
update their code to include a new `isOSMUserId=true` query parameter.
Clients that are exclusively passing MapRoulette user ids are unaffected.

* Add new `isOSMUserId` boolean parameter to project manager API
endpoints that must now be explicitly provided for the given user id(s) to
be treated as an OSM user id(s)

* If `isOSMUserId` is not expicitly set to true in request, treat given
user id(s) as internal MapRoulette ids only

* This new requirement eliminates the potential of erroneously acting on
the wrong target user in the case where a given id matches both internal
and OSM user ids


mgcuthbert:
As a note for the future this will require us to update our minor version at next release.

PR #680, opened 2020-01-24
Add support for challenge snapshots

krotstan:
Provides support for capturing challenge snapshots. A snapshot includes all challenge metrics (how many tasks are in each status), metrics by priority (tasks in each status in a given priority) and review metrics (tasks with a particular review status) from a particular point in time.
 
* Requires evolution - adds new snapshot tables

* Add api calls to record/list/fetch/export snapshots

* Record challenge snapshot on rebuild (and addFileTasks) unless
  asked to 'skipSnapshot'

* ChallengeSummary to also support answered/validated/disabled
  statuses.

* Add job to record snapshots of all challenges. In conf add to
  scheduled jobs:
    challengesSnapshot {
      interval = "7 days"
      startTime = "1:00:00" # Snapshot every week at 1am local time
    }


krotstan:
Done!

PR #684, opened 2020-01-28
Additional support for searching by task id, review requested by and reviewer 

krotstan:
* When fetching task clusters and tasks in bounding box support
  searching by a task id, reviewed by, and review requested by

* When fetching metrics support searching by reviewed by and
  review requested by and task id

* When doing a review search support searching by task id

### DIFF
--- a/app/org/maproulette/data/DataManager.scala
+++ b/app/org/maproulette/data/DataManager.scala
@@ -391,14 +391,13 @@ class DataManager @Inject()(config: Config, db: Database, boundingBoxFinder: Bou
         case None => ""
       }
 
-      var statusFilter = ""
-      var reviewStatusFilter = ""
+      var searchFilters = new StringBuilder
 
       params match {
         case Some(search) =>
           search.taskStatus match {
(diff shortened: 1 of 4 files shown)


app/org/maproulette/data/DataManager.scala
```diff
@@ -411,10 +410,39 @@ class DataManager @Inject()(config: Config, db: Database, boundingBoxFinder: Bou
...
               filter.append(")")
-              reviewStatusFilter = filter.toString()
+              searchFilters.append(filter)
             case Some(statuses) if statuses.isEmpty => //ignore this scenario
             case _ =>
           }
+
+          search.owner match {
+           case Some(o) if o.nonEmpty =>
+             val filter = new StringBuilder(s""" AND (t.id IN (SELECT task_id
```
mgcuthbert: Is there any particular you are creating a StringBuilder object here instead of just putting the string as the parameter in the append function?

krotstan: No good reason, I was just following the prior pattern. I've removed it.

PR #687, opened 2020-02-04
Add support for additional filters when exporting

krotstan:
* When exporting CSV (tasks/extract) support additional
  filters: taskPropertySearch, taskId, reviewer, owner

* When exporting challenge geojson support additional
  filters: taskPropertySearch, taskId, reviwewer, owner

Since taskPropertySearch is a json object, added
api for POST of this data so it can be downloaded with
an HTML Form ```<form>```.


mgcuthbert:
PR looks good, your PR description ended a bit abruptly though.

krotstan:
@mgcuthbert Thanks for catching that! I've updated it.

PR #689, opened 2020-02-06
Delete task_review entry when task set back to created

krotstan:
When setting a task status back to created delete any corresponding task_review entries. This is applicable when rebuildTasks, setStatus, update task.

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -362,22 +362,13 @@ class TaskDAL @Inject()(override val db: Database,
 
       task match {
         case Some(t) =>
-          if (cachedItem.status != t.status) {
-            // If the status is changing and if we have a bundle id, then we need
-            // to clear it out along with any other tasks that also have that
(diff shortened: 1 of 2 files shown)


app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -362,13 +362,22 @@ class TaskDAL @Inject()(override val db: Database,
...
+            // to clear it out along with any other tasks that also have that
+            // bundle id -- essentially breaking up the bundle. Otherwise this
+            // task could end up with a different status than other tasks
+            // in that bundle.
+            if (t.bundleId != None) {
+              this.deleteTaskBundle(user, t.bundleId.get)
+            }
+
+            // If we are moving this task back to a created status, then we don't need to do a review on it.
+            if (t.status.getOrElse(-1) == Task.STATUS_CREATED) {
```
mgcuthbert: I feel like you are doing this twice. The mergeUpdate function already does this anyway, so I am curious if you actually need to execute this delete again.

krotstan: Removed it.

PR #698, opened 2020-02-14
Refactoring

mgcuthbert:
This PR is essentially to reduce the size of the TaskController and TaskDAL files, as the TaskDAL file was getting over 2000 lines long and had some functions that could be logically separated into their own DAL's and Controllers. 

The PR aims to change as little logical code as possible, and for the most part there is very little actual changes other than cosmetic. You will notice the in the end there was only a little over 200 changes. 

It also includes library upgrades specifically Play Framework 2.8.0 (Previously 2.7.0), SBT 1.3.8 (Previously 1.2.4) and Scala 2.13.1 (Previously 2.12.8)

### DIFF
--- a/app/org/maproulette/models/dal/TaskReviewDAL.scala
+++ b/app/org/maproulette/models/dal/TaskReviewDAL.scala
@@ -3,56 +3,91 @@
 package org.maproulette.models.dal
 
 import java.sql.Connection
+import java.time.Instant
 
+import anorm.JodaParameterMetaData._
 import anorm.SqlParser._
 import anorm._
 import javax.inject.{Inject, Provider, Singleton}
-import org.joda.time.{DateTime, DateTimeZone}
+import org.joda.time.DateTime
+import org.maproulette.Config
 import org.maproulette.exception.InvalidException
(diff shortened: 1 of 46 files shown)


app/org/maproulette/models/dal/TaskReviewDAL.scala
```diff
@@ -420,7 +469,7 @@ class TaskReviewDAL @Inject()(override val db: Database,
     val query = user.isSuperUser match {
       case true =>
         s"""
-          SELECT tasks.${this.retrieveColumnsWithReview} FROM tasks
+          SELECT t.${this.retrieveColumnsWithReview} FROM tasks
```
nrotstan: Since you're switching to `t` I think you need `FROM tasks t`

mgcuthbert: Yeah that is a mistake, I switched to t and then realized that was not going to work, obviously I didn't revert every place it happens. Thanks for catching it.

mgcuthbert:
Yeah I need to fix travis compilation before this PR is merged. (So today :) )

PR #728, opened 2020-03-17
Fix time spent calculation to only use db time not server time

krotstan:
If database and server are running on separate machines and
the machines have different time then the time spent calculation
would be off -- change time spent calculations to only use
db times.


krotstan:
Fixed.

PR #743, opened 2020-04-01
Add review status 'unnecessary'

krotstan:
* Add new review status 'unnecessary'

* Add api to 'removeReviewRequest' by marking review status
  unnecessary

* when gathering review metrics exclude reviews unnecessary


mgcuthbert:
test this please

mgcuthbert:
please test this

app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -538,6 +538,51 @@ class TaskController @Inject() (
...
+              params = p.copy(location = Some(SearchLocation(-180, -90, 180, 90)))
+          }
+          val (count, tasks) = this.dalManager.taskCluster.getTasksInBoundingBox(user, params, -1)
+          taskIds = tasks.map(task => task.id)
+        }
+
+        var updatedTasks = 0
+
+        taskIds.foreach(id => {
+          this.dal.retrieveById(id) match {
```
mgcuthbert, line 573: instead of calling each task individually you can do this:
```
// set the taskIds variable to `implicit` above
val updatedTasks = this.dal.retrieveListById.foldLeft(0)(_ + this.dalManager.taskReview.setTaskReviewStatus(t, Task.REVIEW_STATUS_UNNECESSARY, user, None, "")
```

mgcuthbert: Also I didn't check if that compiled, but the general idea should work.

app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -538,6 +538,51 @@ class TaskController @Inject() (
...
+    *
+    * User must have write access to parent challenge(s).
+    *
+    * @param ids The ids of the tasks to update
+    * @return The number of tasks updated.
+    */
+  def removeReviewRequest(ids: String): Action[AnyContent] = Action.async { implicit request =>
+    this.sessionManager.authenticatedRequest { implicit user =>
+      SearchParameters.withSearch { p =>
+        var params = p
```
mgcuthbert, line 553: You should always try to steer clear of `var` in Scala. I have found that in Scala `var` was more a concession that anything else. It is a language that likes to use immutable objects and only switch to mutable objects when absolutely necessary. In general there is rarely a time when you actually need to use `var`.

mgcuthbert:
The build is succeeding, it failed the first time around because Travis was having issues downloading SBT or something so I just reran it and it worked fine. It didn't update the PR though.

app/org/maproulette/controllers/api/TaskController.scala
krotstan: Ok, I've changed it.

app/org/maproulette/controllers/api/TaskController.scala
krotstan: Ok, I've changed it.

PR #755, opened 2020-04-08
Add api to fetch list of challenges for review

krotstan:
Add api '/review/challenges' which will fetch challenges that have tasks that have reviews/review requests. From this list, the front end will be able to narrow down the reviews to be returned to the user by challenge (or even by project) when fetching the full list of reviews.

### DIFF
--- a/app/org/maproulette/controllers/api/TaskReviewController.scala
+++ b/app/org/maproulette/controllers/api/TaskReviewController.scala
@@ -8,7 +8,7 @@ import javax.inject.Inject
 import org.maproulette.Config
 import org.maproulette.data.ActionManager
 import org.maproulette.exception.NotFoundException
-import org.maproulette.framework.model.{Challenge, ChallengeListing, User, Project}
+import org.maproulette.framework.model.{Challenge, User, Project}
 import org.maproulette.framework.service.ServiceManager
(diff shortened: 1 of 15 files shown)


krotstan:
Ok, I've fixed up the query. Please review.

mgcuthbert:
We can merge as soon as the build succeeds. I believe you just need to update the tests as you modified the ServiceManager.

app/org/maproulette/controllers/api/TaskReviewController.scala
```diff
@@ -425,4 +428,77 @@ class TaskReviewController @Inject() (
...
 
+  /**
+    * Returns a list of challenges that have reviews/review requests.
+    *
+    * @param reviewTasksType  The type of reviews (1: To Be Reviewed,  2: User's reviewed Tasks, 3: All reviewed by users)
+    * @param tStatus The task statuses to include
+    * @param excludeOtherReviewers Whether tasks completed by other reviewers should be included
+    * @return JSON challenge list
+    */
+  def listChallenges(
```
mgcuthbert, line 439: This function should go in a new TaskReviewController in the frameworks package so that we can start moving the rest to it eventually.

PR #762, opened 2020-04-16
Add job to expire old review tasks

krotstan:
- Add job to expire old review tasks. To use in conf:
```
     expireTaskReviews.interval="7 days"
     expireTaskReviews.olderThan="180 days"
```
- Fixed setTaskReviewStatus using calculated time instead of
  database time which might cause funny times if server has different
  time than database.

- When fetching reviewed tasks ignore 'unnecessary' tasks in query


app/org/maproulette/framework/service/ChallengeListingService.scala
```diff
@@ -57,6 +57,13 @@ class ChallengeListingService @Inject() (repository: ChallengeListingRepository)
             List(
               // Has a task review
               BaseParameter("task_review.id", "", Operator.NULL, negate = true),
+              // Exclude unnecessary review status
+              BaseParameter(
+                "task_review.review_status",
```
mgcuthbert, line 62: So I merged my implicit tables PR before this one. So you will probably get conflicts as this file was modified. Anyway to handle this table of column is done a little differently, and would be more like this:
```
BaseParameter(
  TaskReview.FIELD_REVIEW_STATUS,
  Task.REVIEW_STATUS_UNNECESSARY,
  negate = true,
  table = Some(TaskReview.TABLE)
)
```

app/org/maproulette/jobs/SchedulerActor.scala
```diff
@@ -196,6 +198,63 @@ class SchedulerActor @Inject() (
...
+    */
+  def expireTaskReviews(str: String): Unit = {
+    config.withFiniteDuration(Config.KEY_SCHEDULER_EXPIRE_TASK_REVIEWS_OLDER_THAN) { duration =>
+      db.withConnection { implicit c =>
+        logger.info(
+          s"Expiring task reviews older than $duration ..."
+        )
+
+        // Note in task review history we are moving these tasks to unnecessary
+        SQL(s"""INSERT INTO task_review_history
```
mgcuthbert, line 214: The scheduler should now being running through the services and in turn the services would call the repositories for any sql. This allows us to limit any sql code in the repositories and try to limit any of it outside of those classes. So you would so something like:
```
service.updateTaskReviewHistory(...)
service.expireTaskReviews(...)
```
Or something like that, and then those service calls would call into the TaskReviewRepository, or TaskReviewHistoryRepository.

app/org/maproulette/models/dal/TaskReviewDAL.scala
```diff
@@ -550,7 +549,8 @@ class TaskReviewDAL @Inject() (
   )(implicit c: Connection = null): (Int, List[Task]) = {
     var orderByClause = ""
     val whereClause = new StringBuilder(
-      "(tasks.bundle_id is NULL OR tasks.is_bundle_primary = true) AND "
+      "(tasks.bundle_id is NULL OR tasks.is_bundle_primary = true) AND " +
```
mgcuthbert, line 552: So one thing I want to start doing is if we modify DAL's and it makes sense we should move the one function that we modify into the new framework, and then add integration tests and postman tests along with it as we move it. This way we can slowly move it piecemeal. Some functions would not really be able to do that as they might be to integrated into other pieces of the code, but it would seem like something like this might be a candidate to do that, although looking at it it seems like quite a long function, so up to you.

app/org/maproulette/framework/service/ChallengeListingService.scala
krotstan: Fixed conflicts.

app/org/maproulette/jobs/SchedulerActor.scala
krotstan: Created a TaskReviewService and moved the sql calls into the TaskReviewRepository. Also, created an initial TaskReviewServiceSpec.

PR #773, opened 2020-04-29
Consolidate groups and user_groups into grants

nrotstan:
* Replace groups and user_groups with grants

* Grants consist of a grantee, a role, and a target object. Grantee and
target object references include both type and id for flexibility

* Stop treating project owners and challenge owners as special cases --
they too must now be granted proper roles to pass permission checks

* In evolution, ensure all existing project owners are granted an admin
role on their projects

* Do not allow last user with admin role to be removed from projects

* API endpoints themselves remain backward compatible, but response
objects will now contain grants instead of groups


nrotstan:
I've added a script that dumps the results of read, write, admin permission checks for every user for every project into a CSV on the file system. The script can be run before and after applying the changes, and then the two CSVs diffed to determine if any permission checks are generating different results. I didn't see any differences in my dev environment.

PR #776, opened 2020-04-30
Clean up empty task tags

krotstan:
When skipping a task and not adding any tags, an empty tag was
being inserted for the task.

### DIFF
--- a/conf/evolutions/default/62.sql
+++ b/conf/evolutions/default/62.sql
@@ -1,6 +1,10 @@
 # --- MapRoulette Scheme
 
 # --- !Ups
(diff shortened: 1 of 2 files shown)


conf/evolutions/default/62.sql
```diff
@@ -0,0 +1,4 @@
+# --- !Ups
+-- Clean up empty task tags.
+-- (This will also clean up the join tables as well. tags_on_tasks/tags_on_challenges)
+DELETE FROM tags where name='';
```
mgcuthbert, line 4: @nrotstan just added a 62.sql evolution, so could you just merge them together.

PR #781, opened 2020-05-04
Add api to delete a snapshot

krotstan:
Add simple api to delete a snapshot by id.

### DIFF
--- a/app/org/maproulette/framework/repository/ChallengeSnapshotResponsitory.scala
+++ b/app/org/maproulette/framework/repository/ChallengeSnapshotResponsitory.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2020 MapRoulette contributors (see CONTRIBUTORS.md).
+ * Licensed under the Apache License, Version 2.0 (see LICENSE).
+ */
+
+package org.maproulette.framework.repository
+
+import java.sql.Connection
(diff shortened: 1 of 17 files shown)


app/org/maproulette/framework/repository/ChallengeSnapshotResponsitory.scala
```diff
@@ -0,0 +1,45 @@
+/*
```
mgcuthbert, line 1: So this file should be named `ChallengeSnapshotRepository` instead of `ChallengeSnapshotResponsitory`

app/org/maproulette/framework/repository/ChallengeSnapshotResponsitory.scala
```diff
@@ -0,0 +1,45 @@
...
+  /**
+    * Deletes a challenge snapshot
+    *
+    * @param snapshotId  The snapshot id to delete
+    */
+  def delete(snapshotId: Long): Unit = {
+    this.withMRTransaction { implicit c =>
+      val query = Query.simple(List())
+
+      query
```
mgcuthbert, line 32: If you are going to use query's for DELETE's, then you should let the query handle your filter, so change too.
```scala
val query = Query.simple(List(BaseParameter(ID, snapshotId)))
query.build(s"DELETE FROM $baseTable").executeUpdate()
```

PR #783, opened 2020-05-08
Add optional includeByTaskStatus to review metrics

krotstan:
* Add to getReviewMetrics optional flag includeByTaskStatus
  that will return current review metrics broken down by
  individual task statuses

* Moved getReviewMetrics method into new framework TaskReviewController
  (taskReviewDAL.getReviewMetrics not converted yet as it relies
   on functionality from TaskDAL and SearchParameters which
   have not been converted to new Query yet)


app/org/maproulette/framework/service/TaskReviewService.scala
```diff
@@ -31,4 +34,41 @@ class TaskReviewService @Inject() (repository: TaskReviewRepository) {
...
+      searchParameters: SearchParameters,
+      mappers: Option[List[String]] = None,
+      reviewers: Option[List[String]] = None,
+      priorities: Option[List[Int]] = None,
+      startDate: String,
+      endDate: String,
+      onlySaved: Boolean = false,
+      excludeOtherReviewers: Boolean = false
+  ): ReviewMetrics = {
+    this.taskReviewDAL.getReviewMetrics(
```
mgcuthbert, line 60: How hard would it be to move this particular function from the DAL to a new repository?

Issue #741, opened 2020-03-30
"No User found using id [-1] to check read access" on attempt to move task to Fixed status


AntonRutkevich:
Request to `PUT task/:id/:status`, eg `PUT /api/v2/task/3742/1?requestReview=true` (move task 3742 to "Fixed" status) produces the following stacktrace:

```
16:27:04.485 [scala-execution-context-global-2191] ERROR o.m.exception.MPExceptionUtil$ - No User found using id [-1] to check read access
org.maproulette.exception.NotFoundException: No User found using id [-1] to check read access
	at org.maproulette.permissions.Permission.hasReadAccess(Permission.scala:67)
	at org.maproulette.models.dal.NotificationDAL.getNotificationSubscriptions(NotificationDAL.scala:207)
	at org.maproulette.models.dal.NotificationDAL.addNotification(NotificationDAL.scala:149)
	at org.maproulette.models.dal.NotificationDAL.createReviewNotification(NotificationDAL.scala:259)
	at org.maproulette.models.dal.TaskDAL.$anonfun$setTaskStatus$5(TaskDAL.scala:677)
	at org.maproulette.models.dal.TaskDAL.$anonfun$setTaskStatus$5$adapted(TaskDAL.scala:609)
	at scala.collection.immutable.List.foreach(List.scala:305)
	at org.maproulette.models.dal.TaskDAL.$anonfun$setTaskStatus$4(TaskDAL.scala:609)
	at org.maproulette.models.utils.TransactionManager.$anonfun$withMRTransaction$1(TransactionManager.scala:29)
	at play.api.db.DefaultDatabase.$anonfun$withTransaction$1(Databases.scala:188)
	at play.api.db.DefaultDatabase.withConnection(Databases.scala:179)
	at play.api.db.DefaultDatabase.withTransaction(Databases.scala:186)
	at org.maproulette.models.utils.TransactionManager.withMRTransaction(TransactionManager.scala:28)
	at org.maproulette.models.utils.TransactionManager.withMRTransaction$(TransactionManager.scala:24)
	at org.maproulette.models.dal.TaskDAL.withMRTransaction(TaskDAL.scala:44)
	at org.maproulette.models.dal.TaskDAL.setTaskStatus(TaskDAL.scala:608)
	at org.maproulette.controllers.api.TaskController.customTaskStatus(TaskController.scala:472)
	at org.maproulette.controllers.api.TaskController.$anonfun$setTaskStatus$2(TaskController.scala:442)
[...]

PR #784, opened 2020-05-11
fix issue 'no user found using id [-1]' if try to change status

sakoumi:
This is fix error "No User found using id [-1] to check read access" on attempt to change task status when review was requested but nobody has reviewed it yet.

In case when new record is created in `task_review` table then `reviewed_by is null` initially. After that if user tries to change task status, error `No User found using id [-1] to check read access` will occur. This is because we are trying to add a notification for an unknown user.


nrotstan:
Note: Fixes #741

PR #789, opened 2020-05-20
Support inverting search parameters

krotstan:
* add new 'invf' SearchParmeter that can include a list of search
  fields that will be inverted (eg. where priority NOT in list)

* rename method "paramsChallengeIds" to "paramsTaskProps"

* changed DataManager.getChallengeSummary to use the
  SearchParametersMixin whenever possible in prepartion
  for conversion to new framework.

* add to DataManager.getChallengeSummary filter by completed_by


krotstan:
> Looks good, one quick question though, does this have Postman tests?

There were no postman tests for TaskReview. I've added some for TaskReview and added an invert fields test.

PR #791, opened 2020-05-27
Fix task completion response 'false' showing as 'als' in CSV

krotstan:


app/org/maproulette/controllers/api/ChallengeController.scala
```diff
@@ -765,7 +765,7 @@ class ChallengeController @Inject() (
                 (responseMap \ key) match {
                   case value: JsDefined =>
                     var propValue = value.get.toString()
-                    if (propValue != "true") {
+                    if (propValue != "true" && propValue != "false") {
                       // Strip off ""s
                       propValue = propValue.substring(1, propValue.length() - 1)
```
mgcuthbert, line 770: Could you elaborate on this exactly. Why do we need to strip off the "s"?

krotstan: It's striping off the quotes (eg. turning "1" into 1).

mgcuthbert: Ah ok, could you update the comment to use quotes instead of the actual quote symbol. I think that is where I got confused.

PR #794, opened 2020-06-01
Add notifications for teams and following

nrotstan:
* Convert NotificationDAL and friends to a service + repos in the new
framework

* Add unit tests around notifications service and repos

* Support two new notification subscription types: team and follow,
which users can use to customize how they receive notifications related
to teams and following, respectively

* Generate notification to a user when they receive an invitation to
join a team

* Generate notification to a user when they get a new follower

* Add support to Query for an optional final clause that is appended to
the very end of the generated SQL to allow things like `RETURNING *` to
be added to the end

* Update some existing unit tests to increase isolation and reduce
potential interactions between tests from shared users, projects, etc.


app/org/maproulette/controllers/api/NotificationController.scala
```diff
@@ -21,7 +22,7 @@ import scala.util.{Failure, Success}
   * @author nrotstan
   */
 class NotificationController @Inject() (
```
mgcuthbert, line 24: You should move the controller into the frameworks package namespace as well.

PR #797, opened 2020-06-04
Add support for fetch nearby review tasks.

krotstan:
* Add new api  /tasks/review/nearby/:proximityId which supports
  passing search parameters to determine which tasks require
  review and are closest to the given proximityId task.

* Also refactored SearchParametersMixin to use new query framework 
  as a first step to moving Tasks and TaskReview into new framework

### DIFF
--- a/app/org/maproulette/framework/mixins/SearchParametersMixin.scala
+++ b/app/org/maproulette/framework/mixins/SearchParametersMixin.scala
@@ -1,0 +1,495 @@
+/*
+ * Copyright (C) 2020 MapRoulette contributors (see CONTRIBUTORS.md).
+ * Licensed under the Apache License, Version 2.0 (see LICENSE).
+ */
+package org.maproulette.framework.mixins
+
+import anorm.NamedParameter
+import org.maproulette.session.{SearchParameters, SearchLocation}
+import org.maproulette.framework.psql.SQLUtils
+import org.maproulette.framework.psql.filter._
+import org.maproulette.framework.psql.{AND, OR}
+import play.api.libs.json.JsDefined
+
+import scala.collection.mutable.ListBuffer
+
+trait SearchParametersMixin {
+
+  /**
+    * Filters by p.display_name with a like %projectSearch%
+    * @param params with inverting on 'ps'
+    */
+  def filterProjectSearch(params: SearchParameters): Parameter[String] = {
+    FilterParameter.conditional(
+      "display_name",
+      s"'${SQLUtils.search(params.projectSearch.getOrElse("").replace("'", "''"))}'",
+      Operator.LIKE,
+      params.invertFields.getOrElse(List()).contains("ps"),
+      true,
+      params.projectSearch != None,
+      Some("p")
+    )
+  }
+
+  /**
+    * Filters by tasks.location with a @ MAKE_ENVELOPE etc.
+    * @param params with inverting on 'tbb'
+    */
+  def filterLocation(params: SearchParameters): Parameter[String] = {
+    this.locationSearch(
+      "location",
+      params.location,
+      "tasks",
+      params.invertFields.getOrElse(List()).contains("tbb")
+    )
+  }
+
+  /**
+    * Filters by c.bounding with a @ MAKE_ENVELOPE etc.
+    * @param params with inverting on 'bb'
+    */
+  def filterBounding(params: SearchParameters): Parameter[String] = {
+    this.locationSearch(
+      "bounding",
+      params.bounding,
+      "c",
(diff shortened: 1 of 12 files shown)


app/org/maproulette/framework/mixins/SearchParametersMixin.scala
```diff
@@ -0,0 +1,495 @@
...
+
+  /**
+    * Filters on tasks task_review.reviewed_by
+    * @param params with inverting on 'r'
+    */
+  def filterReviewer(params: SearchParameters): Parameter[String] = {
+    params.reviewer match {
+      case Some(r) if r.nonEmpty =>
+        val invert = if (params.invertFields.getOrElse(List()).contains("r")) "NOT" else ""
+        CustomParameter(s"""(tasks.id ${invert} IN
```
mgcuthbert, line 400: This can be done a little differently to keep in style with our little DSL, kind of like this:
```scala
SubQueryFilter("tasks.id", Query.simple(
    List(
         BaseParameter("tr.task_id", "tasks.id", useValueDirectly = true), 
         BaseParameter("u.name", s"'%${r}%'", Operator.ILIKE)
    ), 
    "SELECT task_id FROM task_review tr INNER JOIN users u ON u.id = tr.reviewed_by"
)
```

I didn't check the syntax and whether that is exactly correct, but that is the general idea. One thing to watch out for is the table names. There might be an implicit table name being passed down and then appended to all the column names. One thing you might want to update as well is in the Parameter class there is a function called "getColumnKey" which basically sets the column name, it might be useful to check if there is a period in the string and if there is, then don't prepend the provided tableKey.

app/org/maproulette/framework/psql/filter/Operator.scala
```diff
@@ -47,6 +48,7 @@ object Operator extends Enumeration {
       case SIMILAR_TO => s"$negation$key SIMILAR TO $rightValue"
       case EXISTS     => s"${negation}EXISTS ($rightValue)"
       case BOOL       => s"${negation}$key"
+      case AT         => s"${negation}$key @ $rightValue"
```
mgcuthbert, line 51: Nice!

app/org/maproulette/models/dal/TaskReviewDAL.scala
```diff
@@ -375,6 +375,65 @@ class TaskReviewDAL @Inject() (
     }
   }
 
+  /**
+    * Retrieve tasks geographically closest to the given task id wit the
+    * given set of search parameters.
+    */
+  def getNearbyReviewTasks(
```
mgcuthbert, line 382: Any reason why this function is not built in the new framework?

app/org/maproulette/framework/mixins/SearchParametersMixin.scala
krotstan: Updated all three methods that do that subquery.

app/org/maproulette/models/dal/TaskReviewDAL.scala
krotstan: The method is still relying on TaskDal stuff and so would be complicated to convert over. Next step.

mgcuthbert: Ok sounds good.

app/org/maproulette/framework/mixins/SearchParametersMixin.scala
```diff
@@ -0,0 +1,525 @@
...
+
+  /**
+    * Filters by task review status
+    * @param params with inverting on 'trStatus'
+    */
+  def filterTaskReviewStatus(params: SearchParameters): Parameter[String] = {
+    params.taskParams.taskReviewStatus match {
+      case Some(statuses) if statuses.nonEmpty =>
+        val invert = params.invertFields.getOrElse(List()).contains("trStatus")
+        val filter = new StringBuilder(s"""${if (invert) "NOT " else ""}(tasks.id IN
```
mgcuthbert, line 170: This should be converted to a SubQueryFilter

app/org/maproulette/framework/mixins/SearchParametersMixin.scala
```diff
@@ -0,0 +1,525 @@
...
+        .map(t => {
+          SQLUtils.testColumnName(t)
+          s"'${t}'"
+        })
+        .mkString(",")
+
+      val invert = if (params.invertFields.getOrElse(List()).contains("tt")) "NOT " else ""
+
+      CustomParameter(
+        s"""${invert}tasks.id IN (SELECT task_id from tags_on_tasks tt
```
mgcuthbert, line 152: SubQueryFilter.

PR #804, opened 2020-06-23
Add support for tag metrics

krotstan:
* add tag metrics api (metrics grouped by tag given task search criteria):
	/data/tag/metrics

* add tag review metrics api (return metrics grouped by tag given review
  search criteria):
	/tasks/review/tag/metrics

* convert review metrics methods to new framework including:
    - move addSearchToQuery out of DALHelper into searchParametersMixin
    - create ReviewSearchMixin and move method setupReviewSearchClause
      out of TaskReviewDAL

* add tests from new apis and converted methods

* fix bug in UserMetricService where duration was not being properly
  applied

Closes osmlab/maproulette3#1289


app/org/maproulette/framework/mixins/SearchParametersMixin.scala
```diff
@@ -18,45 +18,209 @@ import scala.collection.mutable.ListBuffer
...
 
+  /**
+    * Sets up a where clause searching by given SearchParameters
+    */
+  def filterOnSearchParameters(
+      params: SearchParameters
+  )(implicit projectSearch: Boolean = true): Query = {
+    var newQuery = Query.empty
+
+    newQuery = newQuery.addFilterGroup(this.filterLocation(params))
```
mgcuthbert, line 29: You should probably instead of doing this addFilterGroup over and over again, you should just build the query from scratch like this:
```
val filterList = List(
  this.filterLocation(params),
  this.filterBounding(params),
  ...
)
Query(Filter(filterList))
```
Something like that anyway.

app/org/maproulette/framework/mixins/SearchParametersMixin.scala
```diff
@@ -234,25 +414,29 @@ trait SearchParametersMixin {
             ),
             if (invert) AND() else OR()
           )
-        CustomParameter(s"(${query.sql()})")
-      case Some(statuses) if statuses.isEmpty => CustomParameter("")
-      case _                                  => CustomParameter("")
+        FilterGroup(List(CustomParameter(s"(${query.sql()})")))
```
mgcuthbert, line 417: This kind of seems like a pointless wrapping. The "query" object is actually a FilterGroup, so you could just technically return the object directly.

krotstan: The point of this wrapping is to apply () around the query. Will FilterGroup do that automatically?

PR #807, opened 2020-06-26
Allow tasks to have attachments

nrotstan:
* Allow attachments on task FeatureCollection so that additional data
can be attached to a task, such as reference layers for use during
editing, links to relevant street-level imagery, data for consumption by
external processes, etc.

* Add API endpoint for retrieving the complete JSON representation of a
task attachment, including its raw data and metadata

* Add API endpoint for downloading task attachment data as a file,
decoded if necessary and with the proper mime type (for use with editors
importing reference layers)

* Fix missing `"type": "FeatureCollection"` on task geojson

* Fix typo in TaskRepository.scala filename

* Create TaskRepositorySpec

### DIFF
--- a/app/org/maproulette/framework/controller/TaskController.scala
+++ b/app/org/maproulette/framework/controller/TaskController.scala
@@ -4,6 +4,7 @@
  */
 package org.maproulette.framework.controller
 
+import akka.util.ByteString
 import javax.inject.Inject
 import org.maproulette.data.ActionManager
 import org.maproulette.framework.service.TaskService
@@ -13,6 +14,7 @@ import org.maproulette.session.SessionManager
 import org.maproulette.utils.Utils
 import play.api.mvc._
 import play.api.libs.json._
+import play.api.http.HttpEntity
 import org.maproulette.exception.NotFoundException
 
 /**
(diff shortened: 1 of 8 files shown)


app/org/maproulette/framework/controller/TaskController.scala
```diff
@@ -48,4 +50,67 @@ class TaskController @Inject() (
...
+      }
+  }
+
+  /**
+    * Download the data from a task attachment as a file, decoded if necessary,
+    * and with the correct mime type
+    */
+  def attachmentData(id: Long, attachmentId: String, filename: String): Action[AnyContent] =
+    Action.async { implicit request =>
+      this.sessionManager.userAwareRequest { implicit user =>
```
mgcuthbert, line 73: Shouldn't this be an authenticated session, if you are downloading data to attach to task it probably should have a user associated with that action.

nrotstan: The download request comes straight from JOSM, so it's not possible to send along the proper authentication data.

mgcuthbert: Ah ok, that makes sense.

PR #815, opened 2020-07-08
Add timezone value for exporting csv and geojson

krotstan:
When exporting challenge csv and geoson allow for an optional
timezone offset that will convert mappedOn and reviewedAt dates

### DIFF
--- a/app/org/maproulette/controllers/api/ChallengeController.scala
+++ b/app/org/maproulette/controllers/api/ChallengeController.scala
@@ -652,6 +652,7 @@ class ChallengeController @Inject() (
   ): Action[AnyContent] = Action.async { implicit request =>
     this.sessionManager.authenticatedRequest { implicit user =>
       SearchParameters.withSearch { implicit params =>
+        // Verify timzone offset is valid (eg. -10:00 or +04:00 or 06:30:00)
         val tzRegex = "^[\\+]?([\\-]?[\\d]?\\d)\\:([\\d]?\\d)(\\:\\d\\d)?$".r
         val dateTimeZone =
           timezone match {
--- a/app/org/maproulette/controllers/api/ChallengeController.scala
+++ b/app/org/maproulette/controllers/api/ChallengeController.scala
@@ -11,7 +11,7 @@ import java.util.zip.{ZipEntry, ZipOutputStream}
 import akka.util.ByteString
 import javax.inject.Inject
 import org.apache.commons.lang3.StringUtils
-import org.joda.time.DateTime
+import org.joda.time.{DateTime, DateTimeZone}
 import org.maproulette.Config
 import org.maproulette.controllers.ParentController
 import org.maproulette.data._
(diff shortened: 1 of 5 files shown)


app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -1041,7 +1054,11 @@ class ChallengeDAL @Inject() (
                                             WHEN t.priority = #${Challenge.PRIORITY_MEDIUM} THEN ${Challenge.PRIORITY_MEDIUM_NAME}
                                             WHEN t.priority = #${Challenge.PRIORITY_LOW} THEN ${Challenge.PRIORITY_LOW_NAME}
                                            END)) ||
-                                        hstore('mr_mappedOn', t.mapped_on::text) ||
+                                        hstore('mr_mappedOn',
```
mgcuthbert, line 1057: So I would imagine that the pg_timezone_names table is small, but this request will be made for every single returned row. How many records would this query generally be expected to return? You might want to look at the performance difference of doing a join, instead of a select for each returned record.

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -942,11 +944,22 @@ class ChallengeDAL @Inject() (
...
       priorityFilter: Option[List[Int]] = None,
-      params: Option[SearchParameters] = None
+      params: Option[SearchParameters] = None,
+      timezone: String = Utils.UTC_TIMEZONE
   )(implicit c: Option[Connection] = None): String = {
     this.withMRConnection { implicit c =>
       val filters = new StringBuilder()
 
+      val tzOffset =
+        timezone.matches("^[\\+\\-]*\\d\\d\\:\\d\\d(\\:\\d\\d)?$") match {
```
mgcuthbert, line 954: What is the regex doing? Maybe include a comment to explain.

app/org/maproulette/controllers/api/ChallengeController.scala
```diff
@@ -641,10 +647,23 @@ class ChallengeController @Inject() (
       reviewStatusFilter: String,
       priorityFilter: String,
       filename: String,
-      exportProperties: String = ""
+      exportProperties: String = "",
+      timezone: String
   ): Action[AnyContent] = Action.async { implicit request =>
     this.sessionManager.authenticatedRequest { implicit user =>
       SearchParameters.withSearch { implicit params =>
+        val tzRegex = "^[\\+]?([\\-]?[\\d]?\\d)\\:([\\d]?\\d)(\\:\\d\\d)?$".r
```
mgcuthbert, line 655: Comment to explain this regex would be useful.

app/org/maproulette/models/dal/ChallengeDAL.scala
krotstan: Since the "select pg_timezone_names...." query isn't changing with each row the database optimizer performs that query once and saves the value. I verified this with an "explain" plan on a simplified version of the query.

PR #826, opened 2020-08-05
Allow task status to be changed during review revision

krotstan:
When user is revising a rejected task (or task bundle), allow
  the user to change the task status by passing in a new task status
  to setTaskReviewStatus

### DIFF
--- a/app/org/maproulette/controllers/api/TaskBundleController.scala
+++ b/app/org/maproulette/controllers/api/TaskBundleController.scala
@@ -118,6 +118,7 @@ class TaskBundleController @Inject() (
     * @param reviewStatus The review status id to set the task's review status to
     * @param comment      An optional comment to add to the task
     * @param tags         Optional tags to add to the task
+    * @param newTaskStatus  Optional new taskStatus to change on all tasks in bundle
     * @return 400 BadRequest if task with supplied id not found.
     *         If successful then 200 NoContent
     */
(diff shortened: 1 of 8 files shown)


app/org/maproulette/controllers/api/TaskBundleController.scala
```diff
@@ -118,21 +118,42 @@ class TaskBundleController @Inject() (
...
+      newTaskStatus: String = ""
   ): Action[AnyContent] = Action.async { implicit request =>
     this.sessionManager.authenticatedRequest { implicit user =>
-      val tasks = this.dalManager.taskBundle.getTaskBundle(user, id).tasks match {
+      var tasks = this.dalManager.taskBundle.getTaskBundle(user, id).tasks match {
         case Some(t) => t
         case None    => throw new InvalidException("No tasks found in this bundle.")
       }
 
+      // If the mapper wants to change the task status while revising the task after review
```
mgcuthbert, line 138: You should just move all this code up into the section above by "case Some(t) => ..." 

Then you can keep it as val. If you ever find yourself using a var in Scala you should also look at your code and see if there is a better way to approach it. Scala's methodology moves more towards a functional approach which tries to steer clear of mutable objects.

app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -616,22 +616,39 @@ class TaskController @Inject() (
...
   ): Action[AnyContent] = Action.async { implicit request =>
     this.sessionManager.authenticatedRequest { implicit user =>
-      val task = this.dal.retrieveById(id) match {
+      var task = this.dal.retrieveById(id) match {
         case Some(t) => t
         case None =>
           throw new NotFoundException(s"Task with $id not found, cannot set review status.")
       }
 
+      // If the mapper wants to change the task status while revising the task after review
```
mgcuthbert, line 637: Same comment as above for this section. Also this code in general looks very similar, so may want to see if there is any way to combine it.

app/org/maproulette/models/Task.scala
```diff
@@ -256,26 +256,40 @@ object Task extends CommonField {
...
+    * of review revisions and then these completed statuses can also change to another completed status:
+    * fixed, false_positive, too_hard, already_fixed
+    *
     * @param current The current status of the task
     * @param toSet   The status that the task will be set too
+    * @param allowChange Allow an already completed status to be changed to another completed status.
     * @return True if the status can be set without violating any of the above rules
     */
-  def isValidStatusProgression(current: Int, toSet: Int): Boolean = {
+  def isValidStatusProgression(current: Int, toSet: Int, allowChange: Boolean = false): Boolean = {
```
mgcuthbert, line 268: I don't notice any test updates for this, and if we don't have any tests we definitely should.

krotstan: Added TaskSpec and tests for this method.

app/org/maproulette/controllers/api/TaskController.scala
krotstan: Refactored to remove var

app/org/maproulette/controllers/api/TaskBundleController.scala
krotstan: Refactored to remove var

PR #831, opened 2020-09-02
Add support for multiple custom basemaps

krotstan:
* Create new user_basemaps table and move existing custom
  basemaps into new table

* When updating a user allow for customBasemaps to be specified
  in JSON

* When returning a user return customBasemaps as an array of json
  objects

### DIFF
--- a/app/org/maproulette/framework/repository/UserRepository.scala
+++ b/app/org/maproulette/framework/repository/UserRepository.scala
@@ -102,11 +102,13 @@ class UserRepository @Inject() (
       ewkt: String
   )(implicit c: Option[Connection] = None): User = {
     this.withMRTransaction { implicit c =>
+      updateCustomBasemaps(user)
+
       val query =
         s"""UPDATE users SET name = {name}, description = {description},
                                           avatar_url = {avatarURL}, oauth_token = {token}, oauth_secret = {secret},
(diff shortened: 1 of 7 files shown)


app/org/maproulette/framework/repository/UserRepository.scala
```diff
@@ -330,7 +395,11 @@ object UserRepository {
         ProjectManager(projectId, userId, osmId, displayName, avatarURL.getOrElse(""), roles)
     }
   }
-  private val standardColumns = "*, ST_AsText(users.home_location) AS home"
+  private val standardColumns = "*, ST_AsText(users.home_location) AS home, " +
+    "ARRAY(SELECT id from user_basemaps WHERE user_id=users.id) as custom_basemap_ids, " +
```
mgcuthbert, line 399: So creating a select like this, even on an indexed key, could slow performance. So I would suggested modifying it to be like this: 
```sql
(SELECT ROW(
    ARRAY_AGG(id),
    ARRAY_AGG(name),
    ARRAY_AGG(url),
    ARRAY_AGG(overlay) 
FROM user_basemaps
WHERE user_id = users.id) AS custom_basemaps
```

This will return a record object for that row, so you would need to handle that differently in the parser, but would be a lot more efficient for the call to the database.

mgcuthbert (changes requested):
Just one change requested, which will require a bit of updates to how the results are handled.

mgcuthbert (approved):

PR #838, opened 2020-09-21
Fix mapped_on being cleared on task rebuild

krotstan:
* When tasks are being rebuilt if the task status and mapper are
  not changing then the mapped_on should not change also
    Add to the SQL stored procedure update_task:
```
    --
    -- Only update a null mapped_on if we are resetting task status back to created
    --
    IF task_mapped_on IS NOT NULL THEN
      UPDATE tasks SET mapped_on = task_mapped_on WHERE id = update_id;;
    ELSE
      IF new_status = 0 THEN
        UPDATE tasks SET mapped_on = task_mapped_on WHERE id = update_id;;
      END IF;;
    END IF;;
```

* When tasks are bulk updated, if the status is being set back to
  created we also need to clear the mapper, bundling, form data and review request.

### DIFF
--- a/conf/evolutions/default/69.sql
+++ b/conf/evolutions/default/69.sql
@@ -1,0 +1,167 @@
+# --- !Ups
+
+-- Creates or updates and task. Will also check if task status needs to be updated
+-- This change makes sure not to reset the mapped_on to null when we have an existing task
+DROP FUNCTION IF EXISTS update_task(text,bigint,text,integer,jsonb,jsonb,bigint,integer,bigint,text,timestamp without time zone,integer,integer,integer,timestamp without time zone);;
+CREATE OR REPLACE FUNCTION update_task(task_name text,
+                                        task_parent_id bigint,
+                                        task_instruction text,
+                                        task_status integer,
+                                        geo_json jsonb,
+                                        cooperative_work jsonb DEFAULT NULL,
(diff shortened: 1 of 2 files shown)


conf/evolutions/default/69.sql
```diff
@@ -0,0 +1,167 @@
...
+                             reviewed_by = task_reviewed_by, reviewed_at = task_reviewed_at WHERE task_review.task_id = update_id;;
+    ELSE IF new_status = 0 THEN
+        DELETE FROM task_review WHERE task_review.task_id = update_id;;
+      END IF;;
+    END IF;;
+
+    --
+    -- Only update a null mapped_on if we are resetting task status back to created
+    --
+    IF task_mapped_on IS NOT NULL THEN
```
mgcuthbert, line 78: Seeing as the update sql is the same, you can just use OR here. So 
```sql
IF task_mapped_on IS NOT NULL OR new_status = 0 THEN
...
```

mgcuthbert (changes requested):
One minor request

conf/evolutions/default/69.sql
krotstan: You are right, how silly of me. Fixed!

mgcuthbert (approved):

PR #841, opened 2020-10-01
Add support for returning additional reviewers

krotstan:
* Add to database task_review: 'additional_reviewers' and
  to task_review_history: 'original_reviewer'

* Support 'additionalReviewers' to be returned when fetching
  any task data that includes review data.

* Convert review metrics to fetch from task_review_history to
  better match expected metrics and include metric for
  'additionalReviews'

* Convert reviewer 'leaderboard' to fetch review counts from
  task_review_history and also include how many reviews were
  'additionalReviews'

* When returning review metrics also include breakdown by review status

* Fix bug around updating task_review_history: In setTaskReviewStatus
  the 'started_at' time for the task_review_history should be the
  task's review current 'claimed_at' time. Also include db evolution
  to fix the latest entries for a task in task_review_history by
  populating the started_at from the task_review table.

### DIFF
--- a/app/org/maproulette/controllers/api/TaskController.scala
+++ b/app/org/maproulette/controllers/api/TaskController.scala
@@ -505,9 +505,15 @@ class TaskController @Inject() (
           .toMap
       )
 
+      val allReviewers = new scala.collection.mutable.ListBuffer[Long]()
+      tasks.foreach(t => {
+        allReviewers += t.pointReview.reviewedBy.getOrElse(0L)
+        t.pointReview.additionalReviewers.getOrElse(List()).foreach(r => allReviewers += r)
+      })
+
       val reviewers = Some(
         this.serviceManager.user
-          .retrieveListById(tasks.map(t => t.pointReview.reviewedBy.getOrElse(0L)))
+          .retrieveListById(allReviewers.toList)
           .map(u => u.id -> Json.obj("username" -> u.name, "id" -> u.id))
           .toMap
       )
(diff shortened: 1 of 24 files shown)


app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -505,9 +505,15 @@ class TaskController @Inject() (
           .toMap
       )
 
+      val allReviewers = new scala.collection.mutable.ListBuffer[Long]()
+      tasks.foreach(t => {
```
mgcuthbert, line 509: So a more scala idiosyncratic way of doing this would be to make it not mutable and use a flatMap
```
val reviewers = tasks.flatMap(t => {
    List(t.pointReview.reviewedBy.map(r => r)).flatMap(r => r) ++ t.pointReview.additionalReviewers.getOrElse(List())
})
```

app/org/maproulette/controllers/api/TaskReviewController.scala
```diff
@@ -246,9 +246,15 @@ class TaskReviewController @Inject() (
           .toMap
       )
 
+      val allReviewers = new scala.collection.mutable.ListBuffer[Long]()
```
mgcuthbert, line 249: Same as comment above.

mgcuthbert (changes requested):
One very minor comment

app/org/maproulette/controllers/api/TaskController.scala
krotstan: Thanks! I changed it in both places.

krotstan:
I also added another commit that adds an "additionalReviewer" column to the challenge export CSV

mgcuthbert (approved):

PR #845, opened 2020-10-06
Create GIST index on tasks location column

nrotstan:
* Add evolution ensuring that GIST index is created on tasks `location` column

### DIFF
--- a/conf/evolutions/default/71.sql
+++ b/conf/evolutions/default/71.sql
@@ -1,0 +1,6 @@
+# --- !Ups
+-- Add spatial index to task location, if missing
+CREATE INDEX IF NOT EXISTS idx_tasks_location ON tasks USING GIST (location);;
+
+# --- !Downs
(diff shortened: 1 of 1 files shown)


nrotstan:
One downside to modifying an early evolution is that play servers in development mode will want to execute the downs all the way back and then re-apply the ups, potentially causing data loss for the unwary. I'm happy to resubmit this as a new evolution to avoid that, if desired.

mgcuthbert:
Actually as I looked at the code, I was like "no I don't want to deal with that". So I think we need to make it evolution 71 and I will run the index in the background for this release. And then 71 will run in the next release but because the index already exists it won't do anything.

mgcuthbert (changes requested):
Create index in evolution 71, so that it is available for the next release.

nrotstan:
I've updated this PR to include a new 71.sql evolution instead of modifying the 1.sql evolution

PR #876, opened 2021-02-01
Revert some overly optimistic optimizations

krotstan:
* Revert DataManager to no longer have "with"

* Fix ambigious 'reviewed_at' sort in review queries

* Remove 'with' in TaskClusterRepository and remove inner join on locked

* Remove 'with' in TaskReviewRepository

* Change TaskFilterMixin.add filter locked to use subselect instead of
  join on locked table (as query does not always needed to join on locked)

### DIFF
--- a/app/org/maproulette/framework/mixins/TaskFilterMixin.scala
+++ b/app/org/maproulette/framework/mixins/TaskFilterMixin.scala
@@ -10,6 +10,7 @@ import org.maproulette.framework.service.ServiceManager
 import org.maproulette.framework.model.{Task, User, Challenge, Project}
 import org.maproulette.framework.psql.{Query, _}
 import org.maproulette.framework.psql.filter._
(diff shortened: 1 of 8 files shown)


app/org/maproulette/framework/mixins/TaskFilterMixin.scala
```diff
@@ -75,22 +76,12 @@ trait TaskFilterMixin {
...
-                useValueDirectly = true,
-                table = Some("l")
-              ),
-              BaseParameter(
-                "user_id",
-                user.id,
-                Operator.EQ,
-                useValueDirectly = true,
-                table = Some("l")
+              CustomParameter(
```
mgcuthbert, line 79: This query seems like it would be completely different, not sure if I am just reading it incorrectly.

mgcuthbert (approved):
Minor comment, but looks good.

PR #883, opened 2021-02-09
Fix searching challenges with apostrophe causes SQL error

krotstan:
When searching challenges/mappers/reviewers/projects and
using an apostrophe, the single quote needs to be escaped

### DIFF
--- a/app/org/maproulette/framework/mixins/SearchParametersMixin.scala
+++ b/app/org/maproulette/framework/mixins/SearchParametersMixin.scala
@@ -102,7 +102,7 @@ trait SearchParametersMixin {
       List(
         FilterParameter.conditional(
           Project.FIELD_DISPLAY_NAME,
-          s"'${SQLUtils.search(params.projectSearch.getOrElse("").replace("'", "''"))}'",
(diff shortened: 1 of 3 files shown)


app/org/maproulette/framework/mixins/SearchParametersMixin.scala
```diff
@@ -102,7 +102,7 @@ trait SearchParametersMixin {
       List(
         FilterParameter.conditional(
           Project.FIELD_DISPLAY_NAME,
-          s"'${SQLUtils.search(params.projectSearch.getOrElse("").replace("'", "''"))}'",
+          s"'${SQLUtils.search(params.projectSearch.getOrElse(""))}'",
```
mgcuthbert, line 105: Why was this removed? Is it executed somewhere else? If you don't remove single quotes it will fail in the database.

mgcuthbert (changes requested):
Just need an answer to one question.

app/org/maproulette/framework/mixins/SearchParametersMixin.scala
mgcuthbert: Reading your description, it seems that it would indicate that you should leave it in. I might be misreading it though.

PR #887, opened 2021-04-06
support white space but trim params

jschwarz2030:


app/org/maproulette/controllers/api/ChallengeController.scala
```diff
@@ -755,8 +755,8 @@ class ChallengeController @Inject() (
           challenge.extra.exportableProperties match {
             case Some(ex) =>
               if (!ex.isEmpty) {
-                var propsToExport = ex.replaceAll("\\s", "").split(",")
-                propsToExport.foreach(pe => propsToExportHeaders += pe)
+                var propsToExport = ex.split(",")
+                propsToExport.foreach(pe => propsToExportHeaders += pe.trim)
```
mgcuthbert, line 759: Trim will only remove whitespace characters at the beginning and end of the string. You are changing it such that this will only happen for each element found separated by the comma. I am guessing that most likely this is what was originally intended, are you sure there are no cases that required whitespaces to be removed within the found elements?

mgcuthbert (changes requested):
Just a question.

app/org/maproulette/controllers/api/ChallengeController.scala
mgcuthbert: It seems that this is to support allowing whitespace in the exported column headers. That is what I was missing. If that is the case, then it looks good.

mgcuthbert (approved):

PR #894, opened 2021-05-14
Add changeset_url to challenge schema

jschwarz2030:
resolves https://github.com/osmlab/maproulette3/issues/178 along with this PR: https://github.com/osmlab/maproulette3/pull/1609

### DIFF
--- a/app/org/maproulette/framework/model/Challenge.scala
+++ b/app/org/maproulette/framework/model/Challenge.scala
@@ -100,6 +100,7 @@ case class ChallengeGeneral(
     popularity: Option[Int] = None,
     checkinComment: String = "",
     checkinSource: String = "",
+    changesetUrl: Boolean = false,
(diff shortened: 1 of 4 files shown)


mgcuthbert (changes requested):
Could you explain more how this resolves osmlab/maproulette3#178, I doesn't seem to. Also what exactly is changeset_url referencing? Based on the above issue it would imply that the url is the challenge url and not some changeset. If it was a changeset though which changeset would it exactly be?

jschwarz2030:
I may not be titling the column name very well.  Basically we are adding a boolean to the challenge that will later determine if a url back to the challenge should be constructed and added to the changeset comment when a task is worked on.  The url is built in the FE and will attach it to the comment if changesetUrl is true.

mgcuthbert:
Ok thanks, that makes sense.

mgcuthbert (approved):

PR #899, opened 2021-06-21
add changesetUrl to challenge formatters

jschwarz2030:
In some challenge creation user flows, the frontend is not providing a changesetUrl value of true or false, so we need to add the changesetUrl to the challengeFormatters


mgcuthbert (changes requested):
This type of thing is not done in the ChallengeReads, rather you can do this using the function `updateCreateBody` that you can find in [ChallengeController.scala](https://github.com/maproulette/maproulette2/blob/b4417dedfa7baa479edc7286079f56654715f18d/app/org/maproulette/controllers/api/ChallengeController.scala#L1048) and you can just add a line like this: 

`jsonBody = Utils.insertIntoJson(jsonBody, "changesetUrl", false)(BooleanWrites)`

PR #904, opened 2021-08-02
improve user-friendliness of lock error messages

jschwarz2030:
As a result of https://github.com/osmlab/maproulette3/pull/1631, this PR improves the user-friendliness of the lock messages for task status update services.  I also removed the task ids from the messages for a couple of the services since some tasks are in bundles and don't match up with the task id in the URL, potentially confusing the user.


jschwarz2030:
Going to go ahead and merge this since it's minor changes

PR #930, opened 2021-12-23
update sbt and play plus its dependents; fix compile error in controller

maxgrossman:
After seeing the update to the [docker repo](https://github.com/maproulette/maproulette2-docker/commit/d385a5e60ecef54a6393ccdebb74fae7e7dc9a04) I had some issues getting maproulette compiled. 

These changes get the application compiled and running.

### DIFF
--- a/app/controllers/AuthController.scala
+++ b/app/controllers/AuthController.scala
@@ -62,7 +62,7 @@ class AuthController @Inject() (
           }
         case None =>
           sessionManager.retrieveRequestToken(
-            proxyRedirect(routes.AuthController.authenticate()) + s"?redirect=${getRedirectURL(request, redirect)}"
+            proxyRedirect(routes.AuthController.authenticate) + s"?redirect=${getRedirectURL(request, redirect)}"
           ) match {
             case Right(t) =>
               // We received the unauthorized tokens in the OAuth object - store it before we proceed
--- a/build.sbt
+++ b/build.sbt
@@ -60,8 +60,8 @@ libraryDependencies ++= Seq(
   "commons-codec"           % "commons-codec"       % "1.14",
   "com.typesafe.play"       %% "play-mailer"        % "8.0.0",
   "com.typesafe.play"       %% "play-mailer-guice"  % "8.0.0",
-  "com.typesafe.akka"       %% "akka-cluster-tools" % "2.6.1",
-  "com.typesafe.akka"       %% "akka-cluster-typed" % "2.6.1",
+  "com.typesafe.akka"       %% "akka-cluster-tools" % "2.6.17",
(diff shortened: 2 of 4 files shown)


project/plugins.sbt
```diff
@@ -6,7 +6,7 @@ resolvers ++= Seq(
   Resolver.bintrayIvyRepo("iheartradio", "sbt-plugins")
 )
 
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.0")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.11")
```
maxgrossman, line 9: The update here came after getting the compile error mentioned in this thread [this github thread](https://github.com/sbt/sbt/issues/6400) and following guidance to bump play to >=2.8.8

build.sbt
```diff
@@ -60,8 +60,8 @@ libraryDependencies ++= Seq(
   "commons-codec"           % "commons-codec"       % "1.14",
   "com.typesafe.play"       %% "play-mailer"        % "8.0.0",
   "com.typesafe.play"       %% "play-mailer-guice"  % "8.0.0",
-  "com.typesafe.akka"       %% "akka-cluster-tools" % "2.6.1",
-  "com.typesafe.akka"       %% "akka-cluster-typed" % "2.6.1",
+  "com.typesafe.akka"       %% "akka-cluster-tools" % "2.6.17",
```
maxgrossman, line 63: This came downstream I think from the 2.8.8 update.
I guess the play version introduced .17 for 'core akka' (i assume these are plugins). So then needed these plugins to get bumped as well.

app/controllers/AuthController.scala
```diff
@@ -62,7 +62,7 @@ class AuthController @Inject() (
           }
         case None =>
           sessionManager.retrieveRequestToken(
-            proxyRedirect(routes.AuthController.authenticate()) + s"?redirect=${getRedirectURL(request, redirect)}"
```
maxgrossman, line 65: This one I cannot claim to have full read on. As a scala 'user', not an expert, I understand that parameter-less functions can be referenced/invoked without parentheses. Though can != should so I am not quite sure why a framework bump would enforce this. 

I was getting the compilation error `authenticate does not take parameters`.

maxgrossman:
thanks to log4j, the gift that keeps on giving, need 1.6.1 now. 
https://github.com/sbt/sbt/releases/tag/v1.6.1

ljdelight:
Thank you for the patch, @maxgrossman! Your patch LGTM but there are a few other items that I need to update for the new version (along with more testing), so instead of merging to `dev` we'll get this merged into the `feature-updateSbt` branch (along with this, the new github actions for this project will help verify things are working!).

Thanks again. Merging.

PR #956, opened 2022-05-29
Remove the API_HOST var from swagger.yml

ljdelight:
Since the scala API is what provides the swagger docs there is no need to allow swagger to send requests to a different backend. This patch purges the API_HOST variable and the swagger docs will now, by default, send the request to the host where it (the API serving swagger) is running.

If the API and Docs are separated into different hosts in the future, use this approach https://github.com/iheartradio/play-swagger#how-do-i-use-a-different-host-for-different-environment


ljdelight:
After this is merged, I can merge the associated change in maproulette2-docker deployment scripts https://github.com/maproulette/maproulette2-docker/pull/70

PR #973, opened 2022-09-19
Create the initial CaffeineCache implementation

ljdelight:
This is the initial implementation of a `CaffeineCache` (caffeine is a highly performant in-memory cache) that will soon replace the `BasicCache`. The `BasicCache` was shown to be the limiter in several profiles and lacks a priority queue for items with expiry.

The new dependencies have been added and the code will compile, however, this is NOT REPLACING the `BasicCache` yet. More profiles and tweaks need to be done.

### DIFF
--- a/app/org/maproulette/cache/BasicCache.scala
+++ b/app/org/maproulette/cache/BasicCache.scala
@@ -42,16 +42,7 @@ class BasicCache[Key, Value <: CacheObject[Key]](config: Config) extends Cache[K
     *
     * @return
     */
-  def size: Int = this.cache.size
-
-  /**
-    * True size is a little bit more accurate than size, however the performance will be a bit slower
-    * as this size will loop through all the objects in the cache and expire out any items that have
-    * expired. Thereby giving the true size at the end.
-    *
-    * @return
-    */
-  def trueSize: Int = this.cache.keysIterator.count(!isExpiredByKey(_))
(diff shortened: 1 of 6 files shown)


app/org/maproulette/cache/BasicCache.scala
```diff
@@ -42,16 +42,7 @@ class BasicCache[Key, Value <: CacheObject[Key]](config: Config) extends Cache[K
...
-  def size: Int = this.cache.size
-
-  /**
-    * True size is a little bit more accurate than size, however the performance will be a bit slower
-    * as this size will loop through all the objects in the cache and expire out any items that have
-    * expired. Thereby giving the true size at the end.
-    *
-    * @return
-    */
-  def trueSize: Int = this.cache.keysIterator.count(!isExpiredByKey(_))
```
ljdelight, line 54: `trueSize` isn't used, so purge it.

app/org/maproulette/cache/BasicCache.scala
```diff
@@ -42,16 +42,7 @@ class BasicCache[Key, Value <: CacheObject[Key]](config: Config) extends Cache[K
...
+  def size(): Long = this.cache.size
```
ljdelight, line 45: `size` needs to return a Long to match caffeine.

ljdelight:
Closing to do this with a few more changes and testing.

PR #975, opened 2022-09-21
Add Challenge Comment Notifications for Owners

jschwarz2030:
Owners should be notified when challenge comments are posted, regardless if they're mentioned.

partially resolves https://github.com/osmlab/maproulette3/issues/1822


app/org/maproulette/framework/service/NotificationService.scala
```diff
@@ -192,13 +192,34 @@ class NotificationService @Inject() (
...
+      this.addNotification(
+        UserNotification(
+          -1,
+          userId = challenge.general.owner,
+          notificationType = UserNotification.NOTIFICATION_TYPE_CHALLENGE_COMMENT,
+          fromUsername = Some(fromUser.osmProfile.displayName),
+          taskId = None,
+          challengeId = Some(challenge.id),
+          targetId = Some(comment.id),
+          extra = Some(comment.comment)
```
ljdelight, line 206: What's the end result of this comment in the UI? I ask because if this is a HUGE comment, like a few hundred words, does it make sense to do something like keep only the first 100 or so chars and end it with "..."? Another impact is avoiding sending out a large amount of data unnecessarily.

app/org/maproulette/framework/service/NotificationService.scala
```diff
@@ -426,7 +447,9 @@ class NotificationService @Inject() (
         subscriptions.challengeCompleted
       case UserNotification.NOTIFICATION_TYPE_TEAM   => subscriptions.team
       case UserNotification.NOTIFICATION_TYPE_FOLLOW => subscriptions.follow
-      case _                                         => throw new InvalidException("Invalid notification type")
+      case UserNotification.NOTIFICATION_TYPE_CHALLENGE_COMMENT =>
+        UserNotification.NOTIFICATION_EMAIL_IMMEDIATE
```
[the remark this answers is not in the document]
ljdelight: This is okay, when I commented it was a `1` which was the code smell and why I brought it up.

app/org/maproulette/framework/service/NotificationService.scala
jschwarz2030: I'll discuss message limitations with @mvexel.  this is a good thing to bring up but should be issued separately from this feature and incorporated globally.

PR #976, opened 2022-09-24
Create the CaffeineCache implementation

ljdelight:
This is the implementation of a `CaffeineCache` (caffeine is a highly performant in-memory cache) that can be used based on the conf. The `BasicCache` was shown to be a limiter in several profiles and lacks a priority queue for item expiry.

Use the more performant cache by setting `maproulette.caching.type = "caffeine"` in the conf file. At this time the default will remain the BasicCache until this gets more testing.

### DIFF
--- a/app/org/maproulette/cache/CaffeineCache.scala
+++ b/app/org/maproulette/cache/CaffeineCache.scala
@@ -1,0 +1,139 @@
+package org.maproulette.cache
+
+import com.github.blemale.scaffeine.Scaffeine
+import org.maproulette.Config
+import play.api.Logging
+
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.FiniteDuration
+
+class CaffeineCache[Key, Value <: CacheObject[Key]](config: Config)
+    extends Cache[Key, Value]
+    with Logging {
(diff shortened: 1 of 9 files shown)


app/org/maproulette/cache/CaffeineCache.scala
```diff
@@ -0,0 +1,139 @@
...
+    * @return the current size of the cache
+    */
+  override def size(): Long = caffeineCache.estimatedSize()
+
+  /**
+    * Adds an object to the cache, if cache limit has been reached, then will remove the oldest
+    * accessed item in the cache
+    *
+    * @param key   The object to add to the cache
+    * @param value You can add a custom expiry to a specific element in seconds
```
ben-manes, line 45: fyi, you can use a custom expiration policy, `expireAfter(Expiry)`, if you want per-entry evaluations.

ljdelight: Thanks for the comment and your work on caffeine! It's pretty great :+1:  

Within [maproulette](https://maproulette.org/) the custom expiration doesn't appear to be used (except in a test) and the param should probably be eliminated, so I just ignored it for now

PR #1015, opened 2023-01-23
Revert "Require authN for routes in DataController.scala"

jschwarz2030:
Reverts maproulette/maproulette-backend#1014. These endpoints are needed for non-user public pages.


jschwarz2030:
going to retain this code and decide on different mitigation to resolve nonuser page performance

Issue #1019, opened 2023-02-17
ERROR: CREATE INDEX CONCURRENTLY cannot be executed within a pipeline [ERROR:0, SQLSTATE:25001]


ljdelight:
During startup of a maproulette instance using postgres 13.10, I see this error: `ERROR: CREATE INDEX CONCURRENTLY cannot be executed within a pipeline [ERROR:0, SQLSTATE:25001]`. There is no error when using postgres 13.9.

This was first noticed with CI/CD GitHub actions and is easily repro'd on a local database This appears to be a play framework evolution that doesn't work well with the latest postgres 13 drop.

---
This is one way to start a clean database locally using the latest pg13:
```sh
podman run \
    -d \
    -p 5432:5432 \
    --pull=always \
    --name maproulette-postgis2 \
    --restart unless-stopped \
    --shm-size=2048MB \
    -e POSTGRES_DB=maproulette-db \
    -e POSTGRES_USER=maproulette-db-user \
    -e POSTGRES_PASSWORD=maproulette-db-pass \
    postgis/postgis:13-3.3
```

---

After starting the database and starting MapRoulette within intellij, the error log shows

`22:29:05,771 [error] [play-dev-mode-akka.actor.default-dispatcher-8] p.a.d.e.DatabaseEvolutions.$anonfun$checkEvolutionsState$3(EvolutionsApi.scala:302) - ERROR: CREATE INDEX CONCURRENTLY cannot be executed within a pipeline [ERROR:0, SQLSTATE:25001]`

and after opening the webUI (http://127.0.0.1:9000/docs) this is displayed:
```
We got the following error: ERROR: CREATE INDEX CONCURRENTLY cannot be executed within a pipeline [ERROR:0, SQLSTATE:25001], while trying to run this SQL script:

1  -- Rev:48,Ups - 313b22e
2
3  -- Setup constraints on virtual_project_challenges to restrict duplicate
4  -- entries.
5
6  -- Find and fix prior duplicate entries.
7  DELETE FROM
8  virtual_project_challenges a
9  USING virtual_project_challenges b
10 WHERE
11 a.id < b.id
12 AND a.project_id = b.project_id
13 AND a.challenge_id = b.challenge_id;;
14
15 -- Add unique constraint.
16 CREATE UNIQUE INDEX CONCURRENTLY virtual_project_challenges_projects
17 ON virtual_project_challenges (project_id, challenge_id);;
18
[...]

PR #1022, opened 2023-02-24
remove concurrently statement

jschwarz2030:
Related to https://github.com/maproulette/maproulette-backend/issues/1019


ljdelight:
Indeed this will resolve #1019, see that issue for manual remediation steps for production systems.

PR #1027, opened 2023-03-20
Added Request Review Challenge Setting

CollinBeczak:
Resolves backend for issue:  https://github.com/maproulette/maproulette3/issues/1964

### DIFF
--- a/app/org/maproulette/framework/model/Challenge.scala
+++ b/app/org/maproulette/framework/model/Challenge.scala
@@ -134,7 +134,7 @@ case class ChallengeExtra(
     taskStyles: Option[String] = None,
     taskBundleIdProperty: Option[String] = None,
     isArchived: Boolean = false,
-    reviewSettings: Boolean = false,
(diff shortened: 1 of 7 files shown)


app/org/maproulette/framework/model/Challenge.scala
```diff
@@ -134,6 +134,7 @@ case class ChallengeExtra(
     taskStyles: Option[String] = None,
     taskBundleIdProperty: Option[String] = None,
     isArchived: Boolean = false,
+    reviewSettings: Boolean = false,
```
ljdelight, line 137: The variable `reviewSettings` can be named better. A "review settings" is confusing since a "settings" (general interpretation) can have many options and naming this `reviewSettings` suggests many "settings" and yet it is a boolean. It's misleading to the reader and I thought `boolean` was a typo.

I think better is `tasksNeedReview` or `defaultTaskNeedReview` or `newTasksRequireReview` (is that the best?) or some other name scheme.

PR #1036, opened 2023-04-12
filter out deleted/undiscoverable parents

tsun812:
fix deleted parents filter, add enabled project param filter for tasks cluster endpoints


ljdelight:
Hey @tsun812, know that this PR isn't forgotten. The tests in this area are manual and the patch isn't simple.

PR #1044, opened 2023-05-29
Improve performance of calculating completion percentage in challenges

### LINKED ISSUE #1043: Performance Enhancement Needed for 'updateChallengeCompletionMetrics' Scheduled Job
The task 'updateChallengeCompletionMetrics' is currently set to run every 20 minutes. While it usually finishes within 6-9 minutes, there are instances when it requires more than 30 minutes.

To make this process more efficient, changes need to be made to the `org.maproulette.jobs.SchedulerActor#handleUpdateChallengeCompletionMetrics` method. One suggestion is to bypass the retrieval of the task and instead the SQL query could be adjusted to directly obtain the task counts. Alternatively, we could enhance the SQL query to have postgres calculate the percentage. Implementing either of these changes will significantly decrease the time needed to complete this recurring job.

ljdelight:
This commit changes how the challenge completion percentage is updated. This new approach uses a single SQL query to update all challenges and will complete much faster than the previous.

The new query uses two subqueries to separate the computation of total tasks and completed tasks where, specifically, it makes use of the index on 'tasks.status'.

Resolves #1043 and there's more technical comments there.

### DIFF
--- a/app/org/maproulette/Config.scala
+++ b/app/org/maproulette/Config.scala
@@ -323,8 +323,6 @@ object Config {
     s"$SUB_GROUP_SCHEDULER.archiveChallenges.staleTimeInMonths"
   val KEY_SCHEDULER_UPDATE_CHALLENGE_COMPLETION_INTERVAL =
     s"$SUB_GROUP_SCHEDULER.updateChallengeCompletionMetrics.interval"
-  val KEY_SCHEDULER_UPDATE_CHALLENGE_COMPLETION_START =
-    s"$SUB_GROUP_SCHEDULER.updateChallengeCompletionMetrics.startTime"
   val KEY_SCHEDULER_NOTIFICATION_DIGEST_EMAIL_INTERVAL =
     s"$SUB_GROUP_SCHEDULER.notifications.digestEmail.interval"
   val KEY_SCHEDULER_NOTIFICATION_DIGEST_EMAIL_START =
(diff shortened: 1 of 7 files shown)


ljdelight:
Works fine in staging. Below is a validation sql query that will use the existing 'completion percentage' and compare it to the freshly-computed metric which uses the new code, so if these are different it'll flag an issue. The existing column and generated field are identical, there are no differences in the end result.

```sql
select * from (
SELECT
    challenges.id AS challenge_id,
	challenges.completion_percentage,
    completed_task_counts.completed_tasks * 100 / total_task_counts.total_tasks AS new_completion_percentage
FROM
    challenges
JOIN (
    SELECT
        parent_id,
        COUNT(*) AS total_tasks
    FROM tasks
    GROUP BY parent_id
) AS total_task_counts ON challenges.id = total_task_counts.parent_id
JOIN (
    SELECT
        parent_id,
        COUNT(*) AS completed_tasks
    FROM tasks
    WHERE status != 0
    GROUP BY parent_id
) AS completed_task_counts ON challenges.id = completed_task_counts.parent_id
WHERE NOT deleted and NOT is_archived
) AS challenge_completion_comparison
where challenge_completion_comparison.completion_percentage != challenge_completion_comparison.new_completion_percentage
```

jschwarz2030 (approved):

PR #1059, opened 2023-08-14
update Api Version Number

CollinBeczak:


ljdelight:
Closing, not needed since #1061 sets the version at build time

PR #1060, opened 2023-08-17
Add columns to review table

CollinBeczak:
Add configuration for 3 new columns in the all review related tasks table, and implement them in the review table csv. The columns added are the Feature Ids, challenge ids, and project ids. 
Screen Shot 2023-08-21 at 8 28 09 PM

### DIFF
--- a/app/org/maproulette/framework/controller/TaskReviewController.scala
+++ b/app/org/maproulette/framework/controller/TaskReviewController.scala
@@ -341,10 +341,13 @@ class TaskReviewController @Inject() (
         val csvRows = metrics.map { row =>
           displayedColumns.split(",").flatMap {
             case "Internal Id"   => Seq(row.review.taskId)
+            case "Feature Id"    => Seq(row.task.name)
             case "Review Status" => Seq(Task.reviewStatusMap(row.review.reviewStatus.get))
             case "Mapper"        => Seq(row.review.reviewRequestedByUsername.getOrElse(""))
             case "Challenge"     => Seq(row.review.challengeName.getOrElse(""))
+            case "Challenge Id"  => Seq(row.task.parent)
             case "Project"       => Seq(row.review.projectName.getOrElse(""))
+            case "Project Id"    => Seq(row.review.projectId.getOrElse(""))
             case "Mapped On"     => Seq(row.task.mappedOn.getOrElse(""))
             case "Reviewer"      => Seq(row.review.reviewedByUsername.getOrElse(""))
             case "Reviewed On"   => Seq(row.review.reviewedAt.getOrElse(""))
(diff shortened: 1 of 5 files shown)


app/org/maproulette/framework/mixins/TaskParserMixin.scala
```diff
@@ -149,6 +149,7 @@ trait TaskParserMixin {
       // challenges and projects fields
       get[Option[String]]("challenge_name") ~
       get[Option[String]]("project_name") ~
+      get[Option[Int]]("project_id") ~
```
ljdelight, line 152: There are foreign key constraints on tasks to challenge to project, so the project id will exist and `Option` is not needed here

app/org/maproulette/framework/controller/TaskReviewController.scala
```diff
@@ -341,10 +341,13 @@ class TaskReviewController @Inject() (
         val csvRows = metrics.map { row =>
           displayedColumns.split(",").flatMap {
             case "Internal Id"   => Seq(row.review.taskId)
+            case "Feature Id"    => Seq(row.task.name)
```
ljdelight, line 344: The naming isn't clear to me, is this a feature id or a task name?

app/org/maproulette/framework/controller/TaskReviewController.scala
```diff
@@ -341,10 +341,13 @@ class TaskReviewController @Inject() (
...
             case "Review Status" => Seq(Task.reviewStatusMap(row.review.reviewStatus.get))
             case "Mapper"        => Seq(row.review.reviewRequestedByUsername.getOrElse(""))
             case "Challenge"     => Seq(row.review.challengeName.getOrElse(""))
+            case "Challenge Id"  => Seq(row.task.parent)
```
ljdelight, line 348: In the PR description, would you add the new columns that were added to the csv? An example output (csv headers with one line of data) would be nice to see too.

app/org/maproulette/framework/controller/TaskReviewController.scala
ljdelight: We discussed and a task.name is indeed an osm feature id.

app/org/maproulette/framework/controller/TaskReviewController.scala
ljdelight: Partial screenshot was added, looks good

app/org/maproulette/framework/mixins/TaskParserMixin.scala
ljdelight: reviewHistoryParser creates a TaskReview, and its sql does not include the project id. This should remain an Option.

ljdelight (approved):

PR #1064, opened 2023-09-04
added absent team and follow subscriptions to subscription options

AndrewPhilbin:
Team and Follow options are included in UI settings but were not present in subscription upsert function, making the settings non-functional. Added to upsert function.


ljdelight (approved):
LGTM. Good catch on the missed update, and I checked for any other missing fields and didn't see any.

PR #1065, opened 2023-09-06
Support configurable OSM OAuth2 scope

ljdelight:
This allows for customization of the OSM OAuth2 scope via the conf (osm.oauth2Scope) or an environment variable (MR_OSM_OAUTH2SCOPE). At times the oauth2 scope needs reduced, for example a local dev or staging environment, to avoid accidentally publishing test data to OSM.


ljdelight:
On a test instance I verified that setting `osm.oauth2Scope = "read_prefs"`.

When the write to the user's pref is not allowed (to update the maproulette api key on the user's osm profile), this log message will be generated when logging in:

    INFO  controllers.AuthController - API key unsuccessfully stored in OSM preferences for user id 3. Status code 403

PR #1069, opened 2023-11-03
remove unique name constraint on virtual challenges

CollinBeczak:
resolves: https://github.com/maproulette/maproulette3/issues/2150

### DIFF
--- a/app/org/maproulette/models/dal/ChallengeDAL.scala
+++ b/app/org/maproulette/models/dal/ChallengeDAL.scala
@@ -15,7 +15,7 @@ import org.joda.time.DateTime
 import org.maproulette.Config
 import org.maproulette.cache.CacheManager
 import org.maproulette.data.{Actions, ChallengeType, ProjectType}
-import org.maproulette.exception.{InvalidException, NotFoundException, UniqueViolationException}
+import org.maproulette.exception.{InvalidException, NotFoundException}
 import org.maproulette.framework.model._
 import org.maproulette.framework.repository.{
   ChallengeListingRepository,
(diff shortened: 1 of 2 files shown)


conf/evolutions/default/94.sql
```diff
@@ -0,0 +1,10 @@
+# --- !Ups
+ALTER TABLE virtual_challenges DROP CONSTRAINT IF EXISTS CON_VIRTUAL_CHALLENGES_USER_ID_NAME;
+DROP INDEX IF EXISTS idx_challenges_parent_id_name;
```
ljdelight, line 3: We had discussed removing the index but that's not necessary, remove this line

conf/evolutions/default/94.sql
```diff
@@ -0,0 +1,10 @@
+# --- !Ups
+ALTER TABLE virtual_challenges DROP CONSTRAINT IF EXISTS CON_VIRTUAL_CHALLENGES_USER_ID_NAME;
```
ljdelight, line 2: Add a comment to the file describing the what/why this this is needed. Something like `-- Remove the constraint for a distinct virtual challenge (id,name) tuple; allow users to create virtual challenges with the same name.`

ljdelight (changes requested):

conf/evolutions/default/94.sql
CollinBeczak: Is there a drawback to dropping that index? It's a solution to an error when only dropping the constraint: ERROR: duplicate key value violates unique constraint "idx_challenges_parent_id_name"
  Detail: Key (parent_id, lower(name::text))=(1, tags fix challenge) already exists.

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -513,10 +513,6 @@ class ChallengeDAL @Inject() (
       }
     } match {
```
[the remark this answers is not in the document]
CollinBeczak: Everything functions the same as it did without this fix.

ljdelight (approved):

PR #1094, opened 2023-12-13
Move meta review comments to request body

CollinBeczak:
Frontend: https://github.com/maproulette/maproulette3/pull/2199 && https://github.com/maproulette/maproulette3/pull/2209
Related backend: https://github.com/maproulette/maproulette-backend/pull/1091
Issue fixed: 
Additionally move meta-review comments request body(missed in above related backend PR). The main focus of this fix is so move the rest of the comments to the request body to ensure there will be no more issue with url maximum length being surpassed and to add additional error path handlers to give a better description of whats going wrong when the comment update or creation goes wrong.

Changes:
**Extra important:**
**-----------------------------------------------------------------------------------------**
**The comments were completely removed from these endpoints as they was never needed or updated with this flow.**

Old route to add a comment to challenge
`PUT     /task/:id/:status?comment=string&tags=string`
New route
`PUT     /task/:id/:status?tags=string`

Old route to add a comment to challenge
`PUT     /taskBundle/:bundleId/:status?comment=string&tags=string&errorTags=string`
New route
`PUT     /taskBundle/:bundleId/:status?tags=string&errorTags=string`
**-----------------------------------------------------------------------------------------**

Old route to add a comment to Task
`POST    /task/:taskId/fix/apply?comment=string&tags=string`
New route
`POST    /task/:taskId/fix/apply?tags=string`

Old route to add a comment to challenge
`PUT     /task/:id/metareview/:status?comment=string&tags=string&errorTags=string`
New route
`PUT     /task/:id/metareview/:status?tags=string&errorTags=string`

Old route to add a comment to challenge
`PUT     /task/:id/review/:status?comment=string&tags=string&newTaskStatus=string&newTaskStatus=string&errorTags=string`
New route
`PUT     /task/:id/review/:status?tags=string&newTaskStatus=string&newTaskStatus=string&errorTags=string`

Old route to add a comment to challenge
`PUT     /taskBundle/:id/review/:status?comment=string&tags=string&newTaskStatus=string&newTaskStatus=string&errorTags=string`
New route
`PUT     /taskBundle/:id/review/:status?tags=string&newTaskStatus=string&newTaskStatus=string&errorTags=string`

Old route to add comment to each Task in Bundle
`PUT     /taskBundle/:id/metareview/:status?comment=string&tags=string&errorTags=string`
New route
`PUT     /taskBundle/:id/metareview/:status?tags=string&errorTags=string`

The comments that are now sent through the request body to eliminate the threat of a url character limit being hit.
```
{
  "comment": "This is an example comment."
}
```


app/org/maproulette/framework/controller/TaskBundleController.scala
```diff
@@ -85,14 +83,6 @@ class TaskBundleController @Inject() (
       for (task <- tasks) {
         val action = this.actionManager
           .setAction(Some(user), new TaskItem(task.id), TaskStatusSet(status), task.name)
-        // add comment to each task if any provided
-        if (comment.nonEmpty) {
```
ljdelight, line 89: Isn't this needed since it's adding the provided comment to each task?

app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -731,14 +725,6 @@ class TaskController @Inject() (
 
     val action =
       this.actionManager.setAction(Some(user), new TaskItem(task.id), actionType, task.name)
-    // add comment if any provided
```
ljdelight, line 734: Doesn't this drop the comment if it was provided?

ljdelight (changes requested):

app/org/maproulette/framework/controller/TaskBundleController.scala
ljdelight: After discussion, it appears that this would sync the OSM Comment with the MapRoulette tasks (users would get a MR notification). However this use case hasn't been used in at least 2 years. We're in agreement to remove this.

app/org/maproulette/controllers/api/TaskController.scala
ljdelight: After discussion, it appears that this would sync the OSM Comment with the MapRoulette tasks (users would get a MR notification). However this use case hasn't been used in at least 2 years. We're in agreement to remove this.

ljdelight (approved):

Issue #1102, opened 2024-02-02
Send http error when user is not in leaderboard


ljdelight:
The maproulette leaderboard is a sql table that is updated on some interval. When users send a request for their score, it is fetched from the pre-computed leaderboard in most cases.

However if the userid <:id> is not in the table, it causes the table to be generated in RAM per request. The system is not sufficiently large enough to handle even a few of those requests and the site may fail.

To avoid this, these endpoints must send back errors to the request and NEVER generate the in-memory leaderboard:
- `/data/user/:id/leaderboard`
- `/data/user/:id/topChallenges`
- others?

ljdelight:
Change is live, resolved.

PR #1105, opened 2024-02-23
Improve stability by disabling dynamic leaderboard queries

ljdelight:
Due to the extensive computational demand of dynamic queries for various MapRoulette leaderboards — including those for global standings, project-specific, challenge-specific, and country-specific rankings — system stability issues have been observed.

This patch disables the execution of dynamic leaderboard calculations. Requests triggering a dynamic query now result in an HTTP 400 response.

There was a presumption in the leaderboard code that every user was in the pre-generated leaderboard, and if they were not it would regen the entire thing, for all users, dynamically and per request. So it quickly overwhelmed the DB backend when there were a handful of "new" MapRoulette users to the site. This was especially unfortunate for group mapping parties.

The below leaderboards will not function until a workaround or alternative solution is determined:
- Project-specific leaderboards
- Challenge-specific leaderboards
- Leaderboards featuring items not currently enabled
- Country-specific leaderboards, excluding the global leaderboard, when multiple countries are selected
- Any use of custom date ranges outside the predefined set of durations

Related issues reported by users:
- https://github.com/maproulette/maproulette3/issues/2224
- https://github.com/maproulette/maproulette3/issues/2272
- https://github.com/maproulette/maproulette3/issues/2129

Edit: One leaderboard is so infrequently used that I did not modify its query `/data/reviewer/leaderboard`.

### DIFF
--- a/app/org/maproulette/framework/service/LeaderboardService.scala
+++ b/app/org/maproulette/framework/service/LeaderboardService.scala
@@ -8,7 +8,7 @@ package org.maproulette.framework.service
 import javax.inject.{Inject, Singleton}
 import org.joda.time.DateTime
 import org.maproulette.Config
-import org.maproulette.exception.NotFoundException
+import org.maproulette.exception.{InvalidException, NotFoundException}
 import org.maproulette.framework.model.{LeaderboardChallenge, LeaderboardUser, Task, User}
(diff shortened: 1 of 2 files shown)


ljdelight:
Created to resolve #1102

PR #1106, opened 2024-03-01
add support for relations in overpass queries

CollinBeczak:
When using overpass queries, there are 3 object types: relation, ways, and nodes. Relations are not yet supported, so this PR adds support for relations in overpass queries. Relations are a list of nodes and/or ways and/or relations in one object.


ljdelight:
@CollinBeczak can this be marked as a draft? It's emailing about every commit

ljdelight:
LGTM

ljdelight (approved):

PR #1110, opened 2024-04-09
fix nearby task looping too hard tasks issue

CollinBeczak:
Tickets resolved by this pr: https://github.com/maproulette/maproulette3/issues/2292, https://github.com/maproulette/maproulette3/issues/2018

The reason why users where running into a looping issue with tasks marked as too hard was because the time interval that made it so that tasks marked with mappable status were invisible was based on the value of the "create" column in the task table, which was reset after as a task was set to created or skipped but not too hard since the too hard status is passed through the completion steps instead of the "reset to mappable" steps. To resolve this issue i made the 1 hour interval based on the mapped_on date. I also added some order sorting so that if a user marked a task as too_hard or skipped it that those tasks would be the last option to be selected.

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -1144,7 +1144,7 @@ class TaskDAL @Inject() (
           s"""NOT (tasks.status != ${Task.STATUS_CREATED} AND tasks.id IN (
              |SELECT task_id FROM status_actions
              |WHERE osm_user_id IN (${user.osmProfile.id})
-             |  AND created >= NOW() - '1 hour'::INTERVAL))""".stripMargin
+             |  AND modified >= NOW() - '1 hour'::INTERVAL))""".stripMargin
         )
 
         priority match {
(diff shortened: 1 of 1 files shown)


app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -1144,7 +1144,7 @@ class TaskDAL @Inject() (
           s"""NOT (tasks.status != ${Task.STATUS_CREATED} AND tasks.id IN (
              |SELECT task_id FROM status_actions
              |WHERE osm_user_id IN (${user.osmProfile.id})
-             |  AND created >= NOW() - '1 hour'::INTERVAL))""".stripMargin
```
ljdelight, line 1147: I think there will be unintended consequences by using `modified` vs `created`. The backend has scheduled tasks that run at specific intervals and will update projects/challenges/tasks, and when a task is updated by the automation there is a trigger (`update_tasks()`) which will update the modify time of the task. The `modified` update happens **without user interaction** and is updated any time a scheduled tasks updates a task.

We could have a case where a scheduled task runs, updating the modified date, and those ready-to-be-edited-tasks are excluded from the results until the modified time is over 1 hour. If timing is really unlucky for the scheduled tasks, those could perpetually update the modified time within the hour and, essentially, always filter out all tasks.

CollinBeczak: I'll just add something custom for the too hard status that updates the created column in its normal workflow

ljdelight: Updating the `created` date would cause another set of issues, some other solution is needed.

ljdelight (approved):

PR #1128, opened 2024-06-02
Introduce maproulette.secret.key conf and MAPROULETTE_SECRET_KEY env

ljdelight:
MapRoulette API had a dependency on the Play framework's application key, and this patch separates those keys. The MapRoulette secret is used to encrypt things specific to maproulette and it no longer depends on the Play application key.

Play 2.9 introduced a change where its application key must be 32 bytes or longer, and this impacted MapRoulette's cryptography with its own internal items. It's easier to manage these separately, similar to how the OSM secrets are not used to encrypt data stored within MapRoulette.


ljdelight:
Related to the update to Play 2.9 (#1122) and also the finding by users #1127

PR #1130, opened 2024-06-03
add endpoint for fetching task markers specifically

CollinBeczak:
The ` PUT     /tasks/box/:left/:bottom/:right/:top ` endpoint has some unneeded complexity when fetching for task markers. The reason why that endpoint was so complex is because it is also used to fetch task table data. This led to slower performance when fetching markers because paging, ordering, and sorting was applied.

Because that structure is needed for the tables but not for fetching tasks markers, I have created a new endpoint to reduce complexity: `PUT     /tasks/markers/:left/:bottom/:right/:top `.

This new endpoint exports the same task data as the previous endpoint, but doesn't query for the total task count, and also doesn't do any ordering or paging.

21771 task in view with 1001 task limit:
Old endpoint (Time to complete: ~ 1.1s-1.9s ):
Screenshot 2024-07-19 at 4 36 15 PM

New endpoint  (Time to complete: ~ 400ms-600ms):
Screenshot 2024-07-19 at 4 36 33 PM

977 task in view with 1001 task limit:
Old endpoint  (Time to complete: ~ 450ms-700ms):
Screenshot 2024-07-19 at 4 37 34 PM

New endpoint  (Time to complete: ~300ms-550ms):
Screenshot 2024-07-19 at 4 37 59 PM

Times were recorded on the same device at a similar time, and ran multiple times to get an average. There is a small performance increase with the new endpoint when the number of tasks in view is below the limit. A bigger performance increase is noticeable whenever there are significantly more than 1001 tasks(or the limit) as the endpoint no longer needs to query for all the tasks the meet the filter requirements to get the total tssk count.

### DIFF
--- a/app/org/maproulette/framework/controller/TaskController.scala
+++ b/app/org/maproulette/framework/controller/TaskController.scala
@@ -139,25 +139,22 @@ class TaskController @Inject() (
       top: Double,
       limit: Int,
       excludeLocked: Boolean,
-      includeGeometries: Boolean
+      includeGeometries: Boolean,
+      includeTags: Boolean
   ): Action[AnyContent] = Action.async { implicit request =>
     this.sessionManager.userAwareRequest { implicit user =>
       SearchParameters.withSearch { p =>
         val params = p.copy(location = Some(SearchLocation(left, bottom, right, top)))
-        val (count, result) = this.taskClusterService.getTaskMarkerDataInBoundingBox(
+        val result = this.taskClusterService.getTaskMarkerDataInBoundingBox(
           User.userOrMocked(user),
           params,
           limit,
           excludeLocked
         )
 
-        result match {
-          case Some(points) =>
-            val resultJson = this.insertExtraTaskJSON(points)
-            Ok(Json.obj("total" -> count, "tasks" -> resultJson))
-          case None =>
-            Ok(Json.obj("total" -> count))
-        }
+        val resultJson = this.insertExtraTaskJSON(result, includeGeometries, includeTags)
+
+        Ok(resultJson)
       }
     }
   }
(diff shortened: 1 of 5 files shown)


CollinBeczak:
I'll get this rebased

conf/v2_route/task.api
```diff
@@ -537,6 +537,43 @@ GET     /tasks/random                               @org.maproulette.controllers
...
+#     description: The minimum latitude for the bounding box.
+#   - name: bottom
+#     in: path
+#     description: The minimum longitude for the bounding box.
+#   - name: right
+#     in: path
+#     description: The maximum latitude for the bounding box.
+#   - name: top
+#     in: path
+#     description: The maximum longitude for the bounding box.
```
jake-low, line 563: Nit: these descriptions are mixed up; `left` and `right` are the min and max _longitude_ and `bottom` and `top` are the min and max _latitude_.

This text was copy-pasted from the existing `/tasks/box/:left/:bottom/:right/:top` endpoint and should be fixed there too if you don't mind 🙂

app/org/maproulette/framework/controller/TaskController.scala
```diff
@@ -122,6 +122,43 @@ class TaskController @Inject() (
...
   }
 
+  /**
+    * Gets all the task markers within a bounding box
+    *
+    * @param left   The minimum latitude for the bounding box
+    * @param bottom The minimum longitude for the bounding box
+    * @param right  The maximum latitude for the bounding box
+    * @param top    The maximum longitude for the bounding box
+    * @param limit  Limit for the number of returned tasks
```
jake-low, line 132: The same mixup is also reflected in this doc comment

jake-low (commented):
LGTM other than the noted docs typos, but I think you should wait for @ljdelight's signoff too before merging as he's more familiar with the backend code than I am.

app/org/maproulette/framework/controller/TaskController.scala
CollinBeczak: Thanks for catching this too!

CollinBeczak:
(Approved in direct conversation)

PR #1131, opened 2024-06-04
update tasks_remaining column in challenge table

CollinBeczak:
The scheduled MapRoulette backend task 'updateCompletionMetricsOfActiveChallenges' does not update the `challenges.tasks_remaining` metric, which appears to be its original intent. This field is used in the UI to allow users to sort challenges by the "remaining tasks". 

This PR tweaks the scheduled task to update `challenges.tasks_remaining`, along with the `challenges.completion_progress`.

Resolves: https://github.com/maproulette/maproulette3/issues/2324


ljdelight (approved):
LGTM

PR #1132, opened 2024-06-07
add overlay selector to challenge form backend

tjayschwarz4:
Update to the backend for the ability to select which default overlays will display by default in the challenge form.


CollinBeczak:
Closed because pr was a stale draft pr.
Things that need to be done before opening pull request again:

- [ ] there needs to be a 'comment on column' to help DBAs understand fields without pulling up source
Example: https://github.com/maproulette/maproulette-backend/blob/945625003abf7024186e11f42c85be339aa772ef/conf/evolutions/default/93.sql
- [ ] there needs to be a default, NULL or empty string
- [ ] verify if data is supposed to be in a json format, if so it should be stored as jsonb.

PR #1136, opened 2024-07-15
add global column to challenges and implement filtering for the column

CollinBeczak:
Related Frontend: https://github.com/maproulette/maproulette3/pull/2382

Resolves: https://github.com/maproulette/maproulette3/issues/2373 & https://github.com/maproulette/maproulette3/issues/2350

This PR adds support for a new column in the challenges table called is_global. The challenges.is_global column represents if a challenge is classified as global, currently a challenge is classified as global if it is wider than 180 degrees (half the maps width) or taller than 90 degrees (half the maps height). 

A new .sql file was added to update the global column for challenges that currently exist. This column will also be set when a challenge is created or enters a ready status.


app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -520,6 +524,44 @@ class ChallengeDAL @Inject() (
     }
   }
 
+  def updateSpatialFields(user: User)(implicit challengeId: Long, c: Option[Connection] = None) = {
```
ljdelight, line 527: This is a lot of computation and it will be a concern if this executes on every challenge update (eg a challenge-admin changes the description) or if tasks are added to an existing challenge (eg "bulk task add").

Is the result of the "isGlobal" required immediately or can it wait for a scheduled task to execute? We have a task that updates a challenge's task bounding, and my though is that this computation for "isglobal" could be done in that task (and, in fact, it could look at the challenge's just-computed bounds and not re-compute).

CollinBeczak: either one is an option
>(and, in fact, it could look at the challenge's just-computed bounds and not re-compute).

if we can do that, that would be nice

CollinBeczak:
(Approved in direct conversation)

PR #1137, opened 2024-07-19
fix edge case bug in parsers where avatarUrl is null

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2205

Solution: Make users.avatar_url an option[string] instead of a set string, and then give it an empty string fallback if the column has no value.
I tested this by removing the image of my user and re-rendering the comments:
Screenshot 2024-07-19 at 6 43 26 PM


ljdelight (commented):
Updating a parser works, but there are more in the code that hits the avatar_url. Why not create a migration to have the db require not-null avatar URL and set existing nulls to the current default (/assets/images/user_no_image.png)? That was the preferred in the bug writeup https://github.com/maproulette/maproulette3/issues/2205

PR #1138, opened 2024-07-27
fix stale review_started_at column

CollinBeczak:
Issue: Upon a second attempt to review a task, the review session has the first review session’s start time recorded under the field review_started_at. This results in incorrect, resulting high review time being calculated.

Solution: this pr removes the condition that prevent the column from being updated when a review session is started.

### DIFF
--- a/app/org/maproulette/framework/repository/TaskReviewRepository.scala
+++ b/app/org/maproulette/framework/repository/TaskReviewRepository.scala
@@ -86,7 +86,7 @@ class TaskReviewRepository @Inject() (
         Query
           .simple(List())
           .build(
-             """UPDATE task_review SET review_claimed_by = {userId}, review_claimed_at = NOW(), review_started_at = NOW()
+            """UPDATE task_review SET review_claimed_by = {userId}, review_claimed_at = NOW(), review_started_at = NOW()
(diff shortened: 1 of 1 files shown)


ljdelight (dismissed):
Wouldn't this revert https://github.com/maproulette/maproulette-backend/pull/964 and re-introduce the issue described in that PR?

CollinBeczak:
> Wouldn't this revert #964 and re-introduce the issue described in that PR?

Yes it will, though im not entirely sure that its a problem... The column is used to set the matching column in the task_review_history table. I cant think of a scenario that updating the column when a new reviewer start a task is an issue or how the current implementation doesn't have the same problem since the review time will only get longer because it isnt being update after the first review.

ljdelight:
@CollinBeczak @jschwarz2030, I'll leave it to ya'll to figure out if this is sufficient or adds a regression. The timing itself is not documented very well, neither PR (the previous and this one) details enough to really "know" what the problem was, and there are also zero tests.

jschwarz2030:
The former PR was a preference change on how review time was captured for user review CSV reports.  I am ok with the revert.  It makes more sense that each review gets its own start time as it populated each review in the history table.

jschwarz2030 (approved):

PR #1139, opened 2024-07-27
add `Retrieve a Users Locked Tasks` endpoint

CollinBeczak:
Adds the new endpoint: 
`GET     /user/:userId/lockedTasks?limit={Long}`

This endpoint retrieves a list of data. For each task that is locked by a user, or where the user_id in the locked table matches the requested userId, it returns the task's id, challenge id, challenge name, and the time it was locked.

Example of output:
```
[
    {
        "id": 50109,
        "parent": 63,
        "parentName": "Fix all roads in west Wichita",
        "startedAt": "2024-07-30T17:56:59.650-05:00"
    },
    {
        "id": 50207,
        "parent": 63,
        "parentName": "Fix all roads in west Wichita",
        "startedAt": "2024-07-29T15:40:42.664-05:00"
    },
    {
        "id": 50206,
        "parent": 63,
        "parentName": "Fix all roads in west Wichita",
        "startedAt": "2024-07-29T15:40:42.669-05:00"
    },
    {
        "id": 50210,
        "parent": 63,
        "parentName": "Fix all roads in west Wichita",
        "startedAt": "2024-07-29T15:40:42.673-05:00"
    }
]
```

### DIFF
--- a/app/org/maproulette/framework/model/LockedTask.scala
+++ b/app/org/maproulette/framework/model/LockedTask.scala
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2020 MapRoulette contributors (see CONTRIBUTORS.md).
+ * Licensed under the Apache License, Version 2.0 (see LICENSE).
+ */
+package org.maproulette.framework.model
+
+import org.joda.time.DateTime
+import org.maproulette.cache.CacheObject
+import org.maproulette.framework.psql.CommonField
+import play.api.libs.json.{Json, Format}
+import play.api.libs.json.JodaWrites._
+import play.api.libs.json.JodaReads._
+
+// Define the LockedTask case class
+case class LockedTask(
+    override val id: Long,
+    challengeName: Option[String],
+    startedAt: DateTime
+) extends CacheObject[Long] {
+  // Implement the abstract member 'name'
+  override def name: String = "LockedTask"
+}
+
+/**
+  * Mapping between Task and Challenge and Lock
+  */
+case class LockedTaskData(
+    id: Long,
+    challengeName: Option[String],
+    startedAt: DateTime
+)
+
+object LockedTask extends CommonField {
+  // Use Json.format to automatically derive both Reads and Writes
+  implicit val lockedTaskFormat: Format[LockedTask] = Json.format[LockedTask]
+}
+
+// Define implicit Formats for LockedTaskData
+object LockedTaskData {
+  implicit val lockedTaskDataFormat: Format[LockedTaskData] = Json.format[LockedTaskData]
+}
--- a/app/org/maproulette/framework/model/LockedTask.scala
+++ b/app/org/maproulette/framework/model/LockedTask.scala
@@ -5,35 +5,18 @@
 package org.maproulette.framework.model
 
 import org.joda.time.DateTime
-import org.maproulette.cache.CacheObject
-import org.maproulette.framework.psql.CommonField
 import play.api.libs.json.{Json, Format}
 import play.api.libs.json.JodaWrites._
 import play.api.libs.json.JodaReads._
 
-// Define the LockedTask case class
-case class LockedTask(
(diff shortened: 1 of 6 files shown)


app/org/maproulette/framework/repository/UserSavedObjectsRepository.scala
```diff
@@ -150,6 +151,41 @@ class UserSavedObjectsRepository @Inject() (
...
+        parent     <- get[Long]("tasks.parent_id")
+        parentName <- get[Option[String]]("challenges.challenge_name")
+        lockedTime <- get[DateTime]("locked.locked_time")
+      } yield (LockedTaskData(id, parent, parentName, lockedTime))
+
+      val query = """
+                    SELECT t.id, t.parent_id, l.locked_time, c.name AS challenge_name
+                    FROM tasks t
+                    INNER JOIN locked l ON t.id = l.item_id
+                    LEFT JOIN challenges c ON t.parent_id = c.id
```
ljdelight, line 178: task.parent_id is a FK to challenge.id, so the `LEFT` is misleading and does nothing. Replace it with `INNER`.

conf/v2_route/user.api
```diff
@@ -321,6 +321,30 @@ DELETE   /user/:userId/unsave/:challengeId          @org.maproulette.framework.c
...
+# description: Retrieves a list of all the tasks the user with the matching id has locked
+# responses:
+#   '200':
+#     description: The retrieved Tasks
+#     content:
+#       application/json:
+#         schema:
+#           type: array
+#           items:
+#             $ref: '#/components/schemas/org.maproulette.framework.model.Task'
```
ljdelight, line 334: That's the wrong model, it's not a `Task` but a `LockedTask`

conf/v2_route/user.api
```diff
@@ -321,6 +321,30 @@ DELETE   /user/:userId/unsave/:challengeId          @org.maproulette.framework.c
...
+#   '401':
```
ljdelight, line 335: Nit: where's 404?

conf/v2_route/user.api
```diff
@@ -321,6 +321,30 @@ DELETE   /user/:userId/unsave/:challengeId          @org.maproulette.framework.c
...
+#     description: The user is not authorized to make this request
+# parameters:
+#   - name: userId
+#     in: path
+#     description: The id of the user to retrieve the locked tasks for
+#   - name: limit
+#     in: query
+#     description: Limit the number of results returned in the response. Default value is 50.
```
ljdelight, line 343: Nit: Under description add this bit please
```
          required: false
          schema:
            type: integer
            default: 50
```

app/org/maproulette/framework/model/LockedTask.scala
```diff
@@ -0,0 +1,25 @@
...
+import play.api.libs.json.JodaReads._
+
+/**
+  * Mapping of object structure for fetching task lock data
+  */
+case class LockedTaskData(
+    id: Long,
+    parent: Long,
+    parentName: Option[String],
+    startedAt: DateTime
```
ljdelight, line 19: Please add javadoc explaining what this field indicates. I don't want to require a better name than `startedAt`, but it should be clarified at the very least with javadoc.
Like
```
/**
  * The time that the task was locked
  */
```

ljdelight: Kinda close, the fields themselves is what the javadoc goes on, like

```
/**
  * The time that the task was locked
  */
    startedAt: DateTime
```

conf/v2_route/user.api
ljdelight: >      description: If User or Task for provided ID's is not found.

The wording is wrong, there's no relation to the requesting "User" and a 404. It's only the userId. So keep it to "description: userId is not known" or something like that

ljdelight (approved):

PR #1141, opened 2024-08-11
add users.avatar_url constraint and default

CollinBeczak:
Resolved conversation from https://github.com/maproulette/maproulette-backend/pull/1137
Resolves issue: Resolves: https://github.com/maproulette/maproulette3/issues/2205

This pr adds a new migration file that sets the default of the users.avatar_url column to '/assets/images/user_no_image.png', and prevents its value from being null. This change made it so that avatar_url was no longer an option, so tests and parserss were updated accordingly.

### DIFF
--- a/app/org/maproulette/framework/graphql/schemas/TeamSchema.scala
+++ b/app/org/maproulette/framework/graphql/schemas/TeamSchema.scala
@@ -102,10 +102,7 @@ class TeamSchema @Inject() (override val service: TeamService)
               case Some(description) => Some(description)
               case None              => team.description
             },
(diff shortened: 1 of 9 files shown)


ljdelight (changes requested):
I see unintentional changes to `Groups` and other objects, so I think we should remove about all of the scala changes related to the avatar_url db requirement changes. For example, it's fine to update `user.avatar_url` parsers but no changes should be done for `group.avatar_url` (which looks to always be null).

ljdelight (approved):
LGTM

PR #1144, opened 2024-08-16
Fix edge case in OverpassQL query rewriting

jake-low:
Fixes https://github.com/maproulette/maproulette3/issues/2414

My recent changes to the query rewriting logic had broken the (perfectly valid) query in the linked issue. The problem is that query has settings formatted over multiple lines. I noted in a comment that `rewriteQuery` wouldn't understand these settings, but what I didn't think about was that this would mean it would prepend a second set of settings, and having more than one settings statement is a syntax error in OverpassQL.

The fix I'm proposing here is to check if there appear to be any settings given anywhere in the query, using a crude `String.contains` check that should have no false negatives. If there are, `rewriteQuery` will return the query unmodified, and only if there aren't will it prepend the default `[out:json]` and `timeout` settings.

I'd appreciate a careful review of this. The [list of OverpassQL settings is here](https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Settings) and is what I based the `settings` array in the PR on.

An assumption I'm making with this code that should be reviewed is that an OverpassQL setting starts with `[` + a setting name + `:`, e.g. `[out:json]`. OverpassQL query expressions often also contain `[`, e.g. `way[highway = motorway];`, and they may also contain `:`, e.g. `way["piste:type" = "advanced"];` or `node[amenity = parking](around:100);`. However, **I believe they never contain `[` + a bare string + `:`**, unless that entire sequence appeared inside a quoted string. I'm only searching for a list of known settings names, not `\[.*:`, so the odds of this triggering a false positive seem very low. Therefore, I believe that the check on line 801 **should always be true if the query contains any settings, and additionally should almost never be true if the query does not contain any settings** (something like `node[name ~ "[out:xml]"]` would trigger a false positive but that's a pretty contrived edge case IMO).

Feedback or suggested testcases are welcome.


ljdelight (approved):
LGTM. The items I tested last time still pass, and the user-provided failed queries now pass. In the near future we can simplify this a bit to avoid trying to implement an overpassql parser, but it's good enough to avoid basic issues and improves the errors provided to users.

PR #1149, opened 2024-09-21
Swagger: How to get the API Key and other notes

tordans:
I found it way to hard to collect these bits of information. This is mainly due to the fact that Google links directly to https://maproulette.org/docs/swagger-ui/index.html?url=/assets/swagger.json&docExpansion=none so that becomes the entry point when looking for the API. But also due to the fact that there is no API docs entry page in the docs which could explain some basic things. 

I finally found the info where to get the API key in https://github.com/osmlab/maproulette-python-client/tree/master?tab=readme-ov-file. I suggest to add the info where to find the API key right here; I added them in two places.

I also thing the swagger page should hint at the libraries. I did that too but without linking to them directly so only one place need to keep updates links (the docs page).

Finally, my understanding is that all?? api calls require the API key? Is that true? I think this is important information and I added it to the description as well. Happy to update this with a better explanation on what is public / requires a key.

### DIFF
--- a/conf/swagger.yml
+++ b/conf/swagger.yml
@@ -2,7 +2,7 @@
 openapi: "3.0.3"
 info:
   title: "MapRoulette API"
-  description: "API for MapRoulette enabling the creation and maintenance of MapRoulette challenges"
+  description: "API for MapRoulette enabling the creation and maintenance of MapRoulette challenges. Using the api requires a user apiKey which you can find at the bottom of https://maproulette.org/user/profile. There are Python and Java Wrapper for this API which are linked in the Navigation at https://learn.maproulette.org/en-us/documentation/."
   contact:
     name: "maproulette@maproulette.org"
   license:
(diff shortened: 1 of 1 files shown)


jake-low:
Thanks for making these improvements to the docs; I agree that this seems like a good way to make the required information a bit easier to find.

> Finally, my understanding is that all?? api calls require the API key? Is that true?

Not all API endpoints require a key; I don't have a complete list (and it looks like the Swagger docs don't display this information), but here are some examples of requests that do not require a key:

- `GET https://maproulette.org/api/v2/challenge/48318`
- `GET https://maproulette.org/api/v2/challenge/48318/tasks`
- `GET https://maproulette.org/api/v2/tasks/box/-122.0/47.0/-120.0/49.0`
- `GET https://maproulette.org/api/v2/tasks/random`
- `GET https://maproulette.org/api/v2/projects`
- `GET https://maproulette.org/api/v2/data/user/leaderboard?cLocal=1&limit=10&monthDuration=1&onlyEnabled=true`

These mostly correspond to operations you can do in the UI without being logged in: browsing challenges, viewing the leaderboard, etc.

jake-low (approved):

PR #1155, opened 2024-10-28
Add dockerfile

jake-low:
This PR adds a production-ready Dockerfile. It is similar to what's currently in https://github.com/maproulette/maproulette2-docker, except using a more conventional build workflow (building from the source code in the CWD, instead of using `git clone`).

In a future PR I'll be adding a CI step to build an image from this Dockerfile and push it to GHCR. That image will then be deployed automatically to staging, and when new releases are published, also to prod. Ultimately we won't need the maproulette2-docker repo anymore and can probably archive it.

Having official Docker images will also make it easier for anyone running their own instance of MapRoulette to keep up with new versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+ARG TEMURIN_JDK_TAG=17
+FROM docker.io/library/eclipse-temurin:${TEMURIN_JDK_TAG} AS builder
+
+ARG SBT_VERSION=1.9.9
+
+# This is based on mozilla docker-sbt https://github.com/mozilla/docker-sbt/blob/main/Dockerfile
+# Install sbt
+RUN \
+  curl -L -o sbt-$SBT_VERSION.deb https://repo.scala-sbt.org/scalasbt/debian/sbt-$SBT_VERSION.deb && \
+  dpkg -i sbt-$SBT_VERSION.deb && \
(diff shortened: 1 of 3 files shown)


ljdelight (approved):
lgtm 👍

ljdelight:
> In a future PR I'll be adding a CI step to build an image from this Dockerfile and push it to GHCR. That image will then be deployed automatically to staging, and when new releases are published, also to prod. Ultimately we won't need the maproulette2-docker repo anymore and can probably archive it.

Makes sense to me. The mr2-docker repo may live on if the secrets are kept in the conf files vs within GH as secrets. Depends on how we wanna deploy this beast longer term and how often humans have to get in the way to test a branch (or whatever us humans do and don't want to be inconvenienced by the many args sent to podman)

jake-low:
Yep, agreed. I have started working on a new repo which is a set of OpenTofu (Terraform) configs to deploy the infrastructure, plus post-install scripts to set up the web server reverse proxy and launch the app via Docker Compose. Right now secrets are in a `.env` file but if we're gonna automate that part of the deployment then they'd be GitHub Secrets instead.

The stack is currently running at https://staging-next.maproulette.org/ if you wanna check it out. Not much to see there but it was all deployed automatically. I'll have the deployment scripts cleaned up and published soon for review.

PR #1156, opened 2024-11-05
fix "deletion already in-progress" bug

CollinBeczak:
Issue originally discussed in slack #maproulette-dev channel: https://osmus.slack.com/archives/C04M1NHPEQY/p1730405450254649 

Issue: Challenges are unable to be edited in many ways whenever a task is in the process of deleting tasks. There is an edge case scenario, that this pr fix's, were the "deleting tasks" status on a challenge persists indefinitely.

Steps to reproduce bug bug:
1. Create a challenge with no tasks ( results in a challenge with no status ).
2. Delete tasks in the challenge via the deleteTasks endpoint using the api.
3. Try again and see error.

This is fixed by preventing the challenge status resetting after all tasks are deleted if there was no original status.

### DIFF
--- a/app/org/maproulette/controllers/api/ChallengeController.scala
+++ b/app/org/maproulette/controllers/api/ChallengeController.scala
@@ -578,7 +578,15 @@ class ChallengeController @Inject() (
               try {
                 this.dal.deleteTasks(user, challengeId, Utils.split(statusFilters).map(_.toInt))
               } finally {
-                dalManager.challenge.update(Json.obj("status" -> originalStatus), user)(challengeId)
+                // Check if original status is defined before reverting
+                originalStatus match {
+                  case Some(status) =>
+                    dalManager.challenge.update(Json.obj("status" -> status), user)(challengeId)
+                  case None =>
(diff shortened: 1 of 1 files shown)


ljdelight:
> Steps to reproduce bug bug:
> 
>     1. Create a challenge with no tasks ( results in a challenge with no status ).
>     2. Delete tasks in the challenge via the deleteTasks endpoint using the api.

In step 2, what does it mean to delete tasks from the challenge (which was created with no tasks)? The challenge doesn't have any tasks, and I'm not sure what step 2 is meant to accomplish.

CollinBeczak:
Step 2 just gets the challenge to break, a user is able to attempt to delete tasks in a challenge if it has no tasks. In the discussion were this bug was brought to attention, the user that found the bug did this on accident when they were managing their challenges. Im open to other solutions... we could add a "created" status default to the status column in the challenge table, or somethng else.

CollinBeczak:
Screenshot 2024-11-08 at 2 27 33 PM

PR #1157, opened 2024-11-11
add feature type to csv export

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/1902

This pr add the FeatureType column to the task csv export tables. The main 3 results for the column will be "Point", "LineString", and "GeometryCollection", or unknown if there is no type. Edge case scenarios were another type is used (if this is possible) will results in the type value, example: 

feature: { geometry: [{ type: "MultiLineString" }]} -> FeatureType = "MultilineString". 

Equivalents:
MapRoulette              | OSM
Point                          | Node
LineString                  | Way
GeometryCollection | Relation

### DIFF
--- a/app/org/maproulette/controllers/api/ChallengeController.scala
+++ b/app/org/maproulette/controllers/api/ChallengeController.scala
@@ -824,10 +824,16 @@ class ChallengeController @Inject() (
           }
 
           // Find matching geojson feature properties
-          var propData = ""
+          var propData    = ""
+          var featureType = ""
           task.geojson match {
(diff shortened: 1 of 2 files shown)


app/org/maproulette/controllers/api/ChallengeController.scala
jake-low, line 1: Nice, this is a big improvement in my opinion. Thanks for refactoring this!

jake-low (dismissed):

jschwarz2030 (approved):

PR #1161, opened 2024-12-30
Fix 'sbt: command not found' CI error

jake-low:
This fixes a CI failure that appeared recently. The `ubuntu-latest` runner now refers to Ubuntu 24.04, which does not contain `sbt` by default. Adding the [setup-sbt](https://github.com/sbt/setup-sbt) action resolves the issue.


jake-low:
Merging after discussion at the weekly team meeting

PR #1164, opened 2025-01-13
fix isArchived and isGlobal filtering in extendFind endpoint

CollinBeczak:
**Issue**  
The `/api/v2/challenges/extendedFind` endpoint was not correctly handling archived and global challenge filtering. The issue arose because the filter condition was always evaluating as `false`. This happened because the filter was being wrapped in `Some(value)` (e.g., `Some(false)` or `Some(true)`) instead of directly evaluating the `value`.

**Fix Details**  
This PR updates the condition to handle the wrapped `Some(value)` correctly, ensuring that the filtering logic now behaves as expected. This resolves the bug affecting archived and global challenge filtering.

### DIFF
--- a/app/org/maproulette/framework/mixins/SearchParametersMixin.scala
+++ b/app/org/maproulette/framework/mixins/SearchParametersMixin.scala
@@ -701,7 +701,7 @@ trait SearchParametersMixin {
     */
   def filterChallengeArchived(params: SearchParameters): FilterGroup = {
     params.challengeParams.archived match {
-      case Some(false) =>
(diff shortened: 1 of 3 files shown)


jake-low (approved):
LGTM. This is a hotfix for prod, right? If so, please test on staging after you merge this but before you tag a new release.

CollinBeczak:
Will do, thanks for the review

CollinBeczak:
closing in favor of https://github.com/maproulette/maproulette-backend/pull/1168

PR #1170, opened 2025-02-20
add nearbyTasksWithinBounds endpoints

CollinBeczak:
Needed for https://github.com/maproulette/maproulette3/pull/2552
Adds these two endpoints: 

```
GET     /challenge/:cid/nearby/box/:left/:bottom/:right/:top @org.maproulette.controllers.api.ChallengeController.getNearbyTasksWithinBoundingBox(cid:Long, left:Double, bottom:Double, right:Double, top:Double, limit:Int ?= 5)
```

```
GET     /virtualchallenge/:id/nearby/box/:left/:bottom/:right/:top @org.maproulette.controllers.api.VirtualChallengeController.getNearbyTasksWithinBoundingBox(id:Long, left:Double, bottom:Double, right:Double, top:Double, limit:Int ?= 5)
```

These endpoints allow a user to look at tasks within the set bounds with the same expected results as the other fetchNearbyTasks endpoint.l


app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -1903,4 +1903,45 @@ class ChallengeDAL @Inject() (
...
+            tasks.location && ST_MakeEnvelope($left, $bottom, $right, $top, 4326) AND
+            (l.id IS NULL OR l.user_id = ${user.id}) AND
+            tasks.status IN (0, 3, 6) AND
+            NOT tasks.id IN (
+                SELECT task_id FROM status_actions
+                WHERE osm_user_id = ${user.osmProfile.id} AND created >= NOW() - '1 hour'::INTERVAL)
+      ORDER BY ST_Distance(
+        tasks.location, 
+        ST_SetSRID(ST_MakePoint($centerLon, $centerLat), 4326)
+      ), tasks.status, RANDOM()
```
jake-low, line 1940: What's the reason for `ORDER BY ... RANDOM()`? Seems like tie-breaking on something deterministic (like `tasks.id`) would be preferable since it means that for a given query and DB state the results are deterministic (useful I imagine for unit tests).

jake-low (approved):
LGTM (I checked the bbox parameters and I'm pretty sure they're ordered correctly here)

PR #1189, opened 2025-06-23
Update to sbt 1.11.2 and minor readme updates

ljdelight:


ljdelight:
This appears to be working fine, it's a small update to keep sbt and scalac from getting too old. Merging.

PR #1226, opened 2026-04-09
Add handling for JsNull case in Utils.scala to return empty string

### LINKED ISSUE #2822: Mustache tags null values are returning "null" strings again
### What happened?

I am familiar with this [issue](https://github.com/maproulette/maproulette3/issues/837) which was [resolved](https://github.com/maproulette/maproulette3/pull/1465) back in 2019, but there seems to have been a regression.

I haven't been using it in projects until yesterday, and I found out that  mustache tags encountering null values are returning "null" strings again.

### Steps to reproduce

For example, adding recommended tags in your task instructions, like:

```
`man_made=bridge`
`name=`{{name}}
`alt_name=`{{alt_name}}
```
may return something like, null values return "null" strings.
```
man_made=bridge
name=Xyz Bridge
alt_name=null
```
rather than the previous behavior I expected:
```
man_made=bridge
name=Xyz Bridge
alt_name=
```

### Environment

Firefox,  Vivaldi

### Additional information

_No response_

CollinBeczak:
Resolves: https://github.com/maproulette/maproulette3/issues/2822

### DIFF
--- a/app/org/maproulette/utils/Utils.scala
+++ b/app/org/maproulette/utils/Utils.scala
@@ -51,6 +51,7 @@ object Utils extends DefaultWrites {
           } else {
             try {
               val strValue = kv._2 match {
+                case JsNull      => ""
                 case v: JsNumber => v.toString
                 case v: JsArray  => v.as[Seq[String]].mkString(",")
                 case v           => v.as[String]


jake-low (commented):
Can you provide some more context? This seems like a weird fix for the issue. Utils.getProperties() is called in the Task creation path, so this change affects what gets stored in the database, not just what gets rendered in the frontend.

CollinBeczak:
What i think was happening was when that code block got a null value it would convert it to string value: "null", so i thought that if i just turned it into an empty string that would be fine for now. I suppose i can experiment with having it have a null value and see how the database and frontend handles it, and then make fixes if there are any issues.

ljdelight:
I agree that it shouldn't be fixed this way; it's a generic utility function and would have unintended impact for other callers. 

My suggestion to resolve is (1) reproduce the issue with the mustache syntax being 'null' string and not empty string, (2) use that repro to create a test case in the API project, (3) investigate a solution and fix. A positive is a new test case is created to avoid regression.

PR #1227, opened 2026-04-14
Bypass challenge cache when recalculating task priorities to ensure updated rules and bounds are always applied.

### LINKED ISSUE #2800: Not getting priority tasks from defined high priority bounds.
### What happened?

In this [challenge](https://mpr.lt/c/54564), I've defined a high priority area, where I expect the available tasks to be automatically selected from, if a contributor doesn't select a task manually.

Image

However, this doesn't appear to be the case during my testing, and the same feedback I got from other contributors, so for now, we're resorting to using Virtual Challenges instead.

### Steps to reproduce

1. Open the challenge in Maproullete and tap the "Start" button to get a random task.
2. Verify (from the location box) whether the task is from "Quezon City", a smaller region within the challenge.

### Environment

Linux: Firefox 148, Vivaldi 7.8.3925.76

### Additional information

Occasionally, I would get a task from the defined high priority region but this seems to be rarer than expected, since  I expect most (if not all) randomly selected tasks should be coming from this defined subset of high priority tasks.

CollinBeczak:
resolves: https://github.com/maproulette/maproulette3/issues/2800

### DIFF
--- a/app/org/maproulette/models/dal/ChallengeDAL.scala
+++ b/app/org/maproulette/models/dal/ChallengeDAL.scala
@@ -928,7 +928,9 @@ class ChallengeDAL @Inject() (
   )(implicit id: Long, c: Option[Connection] = None): Unit = {
     this.permission.hasWriteAccess(ChallengeType(), user)
     this.withMRConnection { implicit c =>
-      val challenge = this.retrieveById(id) match {
+      // Bypass the challenge cache so freshly-updated priority rules/bounds are
+      // always used when recalculating task priorities.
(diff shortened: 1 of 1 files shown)


jake-low (approved):
LGTM.

> `Bypass-challenge-cache-when-recalculating-task-priorities-to-ensure-updated-rules-and-bounds-are-always-applied-`

Nit: do you mind using shorter branch names in the future? I check out branches on the CLI often when reviewing and I find really long names like this one a bit cumbersome. Something like `priority-bounds-cache-fix` would be fine IMO.

PR #1228, opened 2026-05-06
Update postgis version used for integration tests

ljdelight:
The GH action for build and integration tests was using postgis:13, this patch updates it to postgis:18-3.6.

Note that production currently uses postgres 17, the unit tests use that version for its action.

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -74,7 +74,7 @@ jobs:
     needs: generate_app_secret
     services:
       postgis:
-        image: postgis/postgis:13-3.5
+        image: postgis/postgis:18-3.6
         ports:
           - 5432:5432
         env:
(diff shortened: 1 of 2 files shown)


jake-low:
We are still running Postgres 17 in production, so I don't think we should drop it from the test matrix. I'm fine with adding Postres 18 in preparation for upgrading production though.

ljdelight:
> We are still running Postgres 17 in production, so I don't think we should drop it from the test matrix. I'm fine with adding Postres 18 in preparation for upgrading production though.

Makes sense and avoids possible issues with patches, I put postgres 17 back in. For postgres 18, yep indeed, adding that in to help ensure the codebase works with the newer version.

ljdelight:
Apologies for the churn, I needed to edit the commit message to be correct since the diff shows that unit tests weren't changed to use a new postgres version. Edited to clarify the actual changes 👍

jake-low (approved):
No worries about the churn. LGTM!

PR #1229, opened 2026-05-25
Update pull-request GH action versions

ljdelight:
Keep the action versions updated to avoid issues with stale versions: checkout v4->v6, setup-java v4->v5. The deployment action file was not updated to avoid issues, it will be updated later.


ljdelight:
The time to build the branch took "3m 32s ", and the PR action happened to match at "3m 32s". 
Both are within the normal build time for this project, no obvious regression, merging.

PR #1233, opened 2026-05-26
Fix priority rules and bounds updates

### LINKED ISSUE #2800: Not getting priority tasks from defined high priority bounds.

CollinBeczak:
Replaces the batched, per-task Scala recompute in `updateTaskPriorities` with a
single SQL `UPDATE … SET priority = CASE … END` issued against the whole
challenge. Priority rules and bounds are translated to SQL once
(`buildPriorityCaseExpression`) and the same expression is reused by
`previewTaskPriorities`, so the editor preview is byte-for-byte identical to
what a save will write. (Preview is used for mr4)

- Pagination loop and 3 follow-up `UPDATE tasks WHERE id IN (…)` per batch
  collapse into one statement, executed in one transaction.
- `WHERE priority IS DISTINCT FROM (…)` skips no-op rows; `RETURNING priority`
  + `GROUP BY` gives the `(high, medium, low)` tally of rows actually changed.
- Polygon bounds → `ST_Intersects(geom, ST_GeomFromGeoJSON(…))`.
- Property rules → `EXISTS (… jsonb_array_elements(features) … jsonb_each_text(properties) …)`
  with case-insensitive key match (mirrors the old Scala
  `equalsIgnoreCase`) and case-sensitive value match.
- Bbox-operator rules (`contains`/`not_contains`) → `location && ST_MakeEnvelope(…, 4326)`.
- fixes issue where only first 1000 tasks were getting the updated priority.

Should resolve https://github.com/maproulette/maproulette3/issues/2800

### DIFF
--- a/app/org/maproulette/models/dal/ChallengeDAL.scala
+++ b/app/org/maproulette/models/dal/ChallengeDAL.scala
@@ -129,8 +129,6 @@ class ChallengeDAL @Inject() (
       get[Int]("challenges.review_setting") ~
       get[Option[String]]("challenges.dataset_url") ~
       get[Option[JsValue]]("challenges.task_widget_layout") ~
-      get[Option[Int]]("challenges.completion_percentage") ~
-      get[Option[Int]]("challenges.tasks_remaining") ~
       get[Boolean]("challenges.require_confirmation") ~
       get[Boolean]("challenges.require_reject_reason") ~
       get[Option[JsValue]]("challenges.completion_metrics") map {
@@ -142,7 +140,9 @@ class ChallengeDAL @Inject() (
             exportableProperties ~ osmIdProperty ~ taskBundleIdProperty ~ preferredTags ~ preferredReviewTags ~
             limitTags ~ limitReviewTags ~ taskStyles ~ lastTaskRefresh ~ dataOriginDate ~ location ~ bounding ~
             requiresLocal ~ deleted ~ isGlobal ~ isArchived ~ reviewSetting ~ datasetUrl ~ taskWidgetLayout ~
-            completionPercentage ~ tasksRemaining ~ requireConfirmation ~ requireRejectReason ~ completionMetricsJson =>
+            requireConfirmation ~ requireRejectReason ~ completionMetricsJson =>
+        val completionMetrics =
+          completionMetricsJson.flatMap(_.asOpt[CompletionMetrics]).getOrElse(CompletionMetrics())
         new Challenge(
           id,
           name,
@@ -210,8 +210,8 @@ class ChallengeDAL @Inject() (
           dataOriginDate,
           location,
           bounding,
-          completionPercentage,
-          completionMetricsJson.flatMap(_.asOpt[CompletionMetrics]).getOrElse(CompletionMetrics())
+          Some(CompletionMetrics.completionPercentage(completionMetrics)),
+          completionMetrics
         )
     }
   }
@@ -279,8 +279,6 @@ class ChallengeDAL @Inject() (
       get[Option[String]]("challenges.dataset_url") ~
       get[Option[JsValue]]("challenges.task_widget_layout") ~
       get[Option[DateTime]]("challenges.system_archived_at") ~
-      get[Option[Int]]("challenges.completion_percentage") ~
-      get[Option[Int]]("challenges.tasks_remaining") ~
       get[Boolean]("challenges.require_confirmation") ~
       get[Boolean]("challenges.require_reject_reason") ~
       get[Option[JsValue]]("challenges.completion_metrics") map {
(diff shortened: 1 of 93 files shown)


app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -1102,75 +1104,238 @@ class ChallengeDAL @Inject() (
...
       }
     }
   }
 
+  // Persists the recomputed priority for every task in the challenge in a single
+  // data-modifying CTE: the UPDATE is filtered by `IS DISTINCT FROM` so rows that
+  // already match the computed priority aren't rewritten, and RETURNING streams
+  // the post-update priority of the rows that actually changed so we can group
+  // them by tier. Returns count of rows written at each priority tier.
+  private def recomputePriorities(
```
jake-low, line 1126: Using `IS DISTINCT FROM` here is the correct choice, since it avoids firing triggers on tasks that were not actually modified. But it also changes the meaning of the tuple returned by `updateTaskPriorities`. I see you called this out in another comment, but no call sites have been updated to handle this change, which makes me wonder: do we need the tuple return value at all? Or should we drop it and make `updateTaskPriorities` return void (or throw on errors)?

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -1102,75 +1104,238 @@ class ChallengeDAL @Inject() (
...
+        // statement text (see method-level note).
+        check.map { c =>
+          s"(geojson IS NOT NULL AND EXISTS (SELECT 1 FROM jsonb_array_elements(" +
+            "CASE WHEN jsonb_typeof(geojson -> 'features') = 'array' " +
+            "THEN geojson -> 'features' ELSE jsonb_build_array() END" +
+            ") AS f, jsonb_each_text(" +
+            "CASE WHEN jsonb_typeof(f -> 'properties') = 'object' " +
+            "THEN f -> 'properties' ELSE jsonb_build_object() END" +
+            s") AS p(k, v) WHERE LOWER(p.k) = LOWER($keyParam) AND $c))"
+        }
```
jake-low, line 1263: This is minor, but I'm pretty sure this logic is slightly different than Task.scala, which does:

```scala
if (parent.isWithinBounds(parent.priority.highPriorityBounds, element) ||
    geometryProps.exists(parent.isHighPriority(_, element))) {
  return Challenge.PRIORITY_HIGH
}

// etc
```

The expression `geometryProps.exists(parent.isHighPriority(_, element)))` returns true if _any_ of the task's features matches the entire condition. If the priority rule is compound (e.g. `k1 = v1 AND k2 = v2`) then the task will only match if one of its features satisfies this condition in full.

OTOH, the SQL translation is dissolving the rule into its parts and checking that each part matches _some_ feature. So if the task has two features, one which satisfies `k1 = v1` and the other which satisfies `k2 = v2`, then the rule as a whole will be considered to match.

This is a really weird edge case so I'm okay with this behavior change even if it might be considered a regression. Tasks with multiple features is already unusual, and I think mainly exists because OSM relations don't map nicely onto GeoJSON features, and are sometimes modeled as FeatureCollections. It's not obvious what the right interpretation of a priority rule is when applied to such a FeatureCollection. We have similarly weird code on the frontend to deal with multi-feature tasks; for example when we build the task properties dict (to use when evaluating mustache templates or displaying the properties in the UI) I think we `Object.assign()` all of the feature properties together in some code paths, and take `features[0].properties` in others.

So anyway, I'm fine with this change but just wanted to call it out in case you think it might cause problems that the challenge logic and the task logic may disagree in rare cases.

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -1102,75 +1104,238 @@ class ChallengeDAL @Inject() (
...
+  )(implicit id: Long, c: Connection): Map[Int, Long] = {
+    val (expr, params) = buildPriorityCaseExpression(challenge)
+    val allParams      = params :+ NamedParameter("pid", id)
+    SQL(
+      s"""WITH updated AS (
+            UPDATE tasks SET priority = $expr
+            WHERE parent_id = {pid} AND priority IS DISTINCT FROM ($expr)
+            RETURNING priority
+          )
+          SELECT priority, COUNT(*) AS cnt FROM updated GROUP BY priority"""
```
jake-low: You can avoid double-computing the `expr` here with a second CTE:

```sql
﻿﻿WITH computed AS (
  SELECT id, ($expr) AS priority FROM tasks WHERE parent_id = {pid}
), updated AS (
  UPDATE tasks
  SET priority = computed.priority FROM computed
  WHERE tasks.id = computed.id
    AND tasks.priority IS DISTINCT FROM computed.priority
  RETURNING t.priority
)
SELECT priority, COUNT(*) FROM updated GROUP BY priority
```

jake-low (dismissed):
This looks like a nice improvement. I left a couple suggestions but overall this LGTM. If it's a hassle to make changes and retest this then I'm okay merging as-is.

app/org/maproulette/models/dal/ChallengeDAL.scala
CollinBeczak: k, got it updated

app/org/maproulette/models/dal/ChallengeDAL.scala
CollinBeczak: Added a note about it

app/org/maproulette/models/dal/ChallengeDAL.scala
CollinBeczak: UpdateTaskPriorities now returns Unit, controller surfaces tasksByPriority post-counts instead.

PR #1236, opened 2026-05-27
add support for maproulette4 endpoints

CollinBeczak:
# Deploying after "add support for new features" branch in initially merged

This branch adds a batch of backend features and a handful of database
migrations. A few of those migrations don't take full effect on their own, so
there are some manual steps to run when this goes out. This document walks
through what changed, what you have to do by hand, and what to keep an eye on
afterward.

## What's in this branch

- **A new tile system for the explore map.** The map's task clusters are now
  served from a pre-computed pyramid of grid cells instead of being calculated
  on the fly. The database keeps these cells up to date automatically: whenever
  tasks or challenges change, a trigger flags the affected cells, and a
  background worker rebuilds them within about a second. The highest zoom level
  is still drawn live from the tasks table.

- **Completion metrics stored on challenges and projects.** Each challenge and
  project now carries a small summary of its task statuses (total, available,
  fixed, skipped, and so on) so that figure is available without a separate
  stats query. The numbers are filled in during the migration and kept current
  by database triggers from then on.

- **Priority bounds on challenges enhancements.** 

- **Skip and archive support for tasks.** 

- **Smaller additions:** trigram-based task name search, challenge "likes," a
  per-user plugins setting, a Nominatim lookup service, and various new API
  routes. The scheduled job that recomputes a challenge's bounding box also uses
  a slightly different, lighter calculation now.

### Step 1 — Force the migrations to re-run

The migrations live in the `play_evolutions` table. Migration 106 was rewritten
and migrations 107 through 112 are new, so we want Play to apply all of them
fresh. Before restarting the app, delete the tracking rows for everything past
105:

```sql
DELETE FROM play_evolutions WHERE id > 105;
```

When the app restarts it will see those migrations as "not yet applied" and run
them forward. It will **not** run the rollback half of each migration, which is
what we want — the rollbacks are destructive (for example, the tile system's
rollback drops all of it). The forward halves are written to be safe to run
even if some of their objects already exist.

> **Note about migration 105:** 105 itself was touched (a whitespace-only
> change), so its checksum no longer matches what's recorded. If Play refuses to
> start and points at 105, delete its row too (`WHERE id >= 105`). Re-running
> 105 is harmless — it just rebuilds an index.

### Step 2 — Make sure the trigram extension can be installed

Migration 108 turns on the PostgreSQL `pg_trgm` extension (used for task-name
search). Installing an extension needs elevated database privileges. If the
application's database user can't create extensions, have a superuser run this
once ahead of time so migration 108 doesn't fail:

```sql
CREATE EXTENSION IF NOT EXISTS pg_trgm;
```

### Step 3 — Build the initial map tiles

The tile system's storage table is created **empty**. Until it's populated, the
explore map will show no clusters at the lower zoom levels. After the migrations
have applied, run this once to build the whole pyramid:

```sql
SELECT rebuild_all_tile_cells();
```

This scans every task, so it's a heavy one-time operation — run it during a
quiet period. From then on the background worker keeps everything current, so
you won't need to run it again under normal circumstances.

## What to test and watch afterward

- **Priority rules and bounds.** 

- **Skipping tasks.** 

- **Tile system performance under bulk changes.** Every change to a task now
  fires two triggers (one for the tiles, one for completion metrics). That's
  cheap for a single edit but adds up during bulk operations — challenge
  refreshes, large status updates, deletes, and the scheduled jobs that update
  challenge locations and snapshots. Keep an eye on how those bulk operations
  perform, on the size of the tile worker's backlog after a big refresh, and on
  how long the initial `rebuild_all_tile_cells()` takes.


CollinBeczak:
I tested these changes locally with maproulette3 frontend and it runs fine. My biggest concern at the moment is the scheduled actions.

app/org/maproulette/controllers/api/ChallengeController.scala
```diff
@@ -121,6 +121,23 @@ class ChallengeController @Inject() (
...
+    */
+  override def read(implicit id: Long): Action[AnyContent] = Action.async { implicit request =>
+    this.sessionManager.userAwareRequest { implicit user =>
+      implicit val baseChallengeWrites = BaseChallenge.baseChallengeWrites
+      this.dal.retrieveBaseChallengeById match {
+        case Some(value) => Ok(Json.toJson(value))
+        case None        => NotFound
+      }
+    }
+  }
```
jake-low, line 139: This is a breaking change to the API (`GET /challenge/:id` will now return data in a new, flatter JSON shape). I'm not sure if this is a good idea.

AFAICT the changes are:
1. flattened JSON structure, e.g. `challenge.general.owner` becomes `challenge.owner`
2. `location` and `bounding` are now nested GeoJSON values instead of strings containing stringified JSON
3. `tasksRemaining` has been replaced by `completionMetrics` (from migration 111)

These all seem like fine changes in isolation, but my concern is that:
1. these will definitely break maproulette3, and likely also any third-party consumers of this API endpoint
2. the change seems incomplete; the old `Challenge.scala` model exists and is still returned by the list, find, and extendedFind endpoints, and is still expected by the `PUT /challenge/:id` endpoint when updating a challenge.

I think if we make this change, we should do so deliberately across all endpoints (and will need to update maproulette3 code to handle the change before deploying). A simpler alternative might be to just do (2) and (3) above, but not (1), which limits the frontend code that will need to be modified.

app/org/maproulette/models/utils/ChallengeFormatters.scala
```diff
@@ -108,6 +110,23 @@ trait ChallengeWrites extends DefaultWrites {
...
+      case Some(pct) if pct > 0 && pct < 100 =>
+        Math.round(remaining.toDouble / (1.0 - pct / 100.0)).toInt
+      case _ => remaining
+    }
+    val completionMetrics = Json.toJson(
+      CompletionMetrics(total = total, available = remaining, tasksRemaining = remaining)
+    )
+    base + ("completionMetrics" -> completionMetrics)
+  }
 }
```
jake-low: I think this means that challenges returned by list/find/extendedFind (which use the Challenge model) will have incomplete/wrong stats, whereas challenges returned by `GET /challenge/:id` will have accurate ones. I don't think the per-status stats are used in the maproulette4 frontend (e.g. on the explore page, where several challenges are visible at once, the progress bar just shows total % completion, but does not break down by status). So I guess this is okay? But it would be better to omit the incorrect stats (return a different JSON structure with fewer fields) rather than providing wrong stats. Or better yet, read the actual `completion_metrics` JSONB column in the legacy Challenge DAL, so that all endpoints always return complete and accurate challenge metrics.

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -915,17 +1089,37 @@ class ChallengeDAL @Inject() (
...
+    * Will run through the tasks in batches and recompute each task's priority from the
+    * challenge's current rules and bounds. Returns a `(high, medium, low)` tuple of the
+    * number of task rows written at each level — callers (particularly the priorities
+    * endpoint) use this as an honest receipt that the DB actually changed, since a silent
+    * 0/0/0 is a strong signal something upstream is wrong.
     *
     * @param user The user executing the request
     * @param id   The id of the challenge
     * @param c    The connection for the request
     */
```
jake-low: I think this comment is meant to describe `updateTaskPriorities`; it's in the wrong place.

Also, I'm confused by the comment's claim that callers need to look at the number of affected rows returned by the function in order to tell whether it succeeded or failed. Is there a reason we can't raise an exception on failure, instead of leaving callers to guess? (Returning the number of affected rows is fine, but it shouldn't be the only signal of whether the function succeeded).

app/org/maproulette/jobs/SchedulerActor.scala
```diff
@@ -168,11 +168,11 @@ class SchedulerActor @Inject() (
         db.withTransaction {
           implicit c =>
             val query =
-              s"""UPDATE challenges 
+              s"""UPDATE challenges
                   SET location = (SELECT ST_Centroid(ST_Collect(ST_Makevalid(location)))
                                  FROM tasks
                                  WHERE parent_id = ${id}),
-                      bounding = (SELECT ST_Envelope(ST_Buffer((ST_SetSRID(ST_Extent(location), 4326))::geography,2)::geometry)
+                      bounding = (SELECT ST_Envelope(ST_Expand(ST_SetSRID(ST_Extent(location), 4326), 0.0001))
```
jake-low: This change looks reasonable (ST_Expand is faster than casting to geography and bufferring in spherical coordinates, and the result is similar). But I think we can go further: why should `bounding` be buffered at all? It's faster and more correct to just have `bounding` be the exact `ST_Extent` of the task's geometry (`location`). If the geometry is a point, this will be a degenerate bounding box, but that's still valid and can still be tested for intersection using `&&` with other bounding boxes (e.g. to get all tasks within the viewport for the explore page).

conf/evolutions/default/105.sql
```diff
@@ -5,7 +5,7 @@
 -- This allows challenge names to be reused after deletion
 DROP INDEX IF EXISTS idx_challenges_parent_id_name;;
 
-CREATE UNIQUE INDEX idx_challenges_parent_id_name ON challenges (parent_id, lower(name)) 
+CREATE UNIQUE INDEX idx_challenges_parent_id_name ON challenges (parent_id, lower(name))
```
jake-low: No need to change this, but for future reference it's best not to modify previously applied migrations (even to correct typos or formatting errors), since Play tracks the checksum of each migration and will refuse to apply new migrations if it can't match its recorded history with the code in the repo. The workaround is to manually modify the `play_evolutions` table as noted in the PR description.

conf/evolutions/default/106.sql
```diff
@@ -1,6 +1,5 @@
 # --- !Ups
-ALTER TABLE users ADD COLUMN show_priority_marker_colors BOOLEAN DEFAULT false;;
+ALTER TABLE users ADD COLUMN IF NOT EXISTS show_priority_marker_colors BOOLEAN DEFAULT false;;
```
jake-low: Likewise, I don't think this modification is strictly necessary (older versions of the `users` table should not have this column; if they do, it indicates that the database schema expected by this migration differs from what's in production, which is an error that should be resolved manually instead of automatically bypassed in the migration code.

app/org/maproulette/jobs/TileDirtyListener.scala
```diff
@@ -0,0 +1,151 @@
...
+      var total   = 0
+      var batches = 0
+      var n       = service.rebuildDirtyCells(DrainBatch)
+      total += n
+      while (n >= DrainBatch && batches < MaxBatchesPerWake) {
+        n = service.rebuildDirtyCells(DrainBatch)
+        total += n
+        batches += 1
+      }
+      if (total > 0) logger.info(s"Tile listener rebuilt $total dirty leaf cells")
```
jake-low: We should also log here if `batches == MaxBatchesPerWake` (so that if the worker is consistently busy and can't keep up with the rate at which tile cells are being marked dirty, we'll see evidence for that in the logs).

conf/evolutions/default/107.sql
```diff
@@ -0,0 +1,458 @@
...
+        lat_to_tile_y(ST_Y(t.location), 15)
+    FROM tasks t
+    WHERE t.parent_id = NEW.id
+      AND t.location IS NOT NULL
+      AND NOT ST_IsEmpty(t.location)
+      AND ST_X(t.location) BETWEEN -180 AND 180
+      AND ST_Y(t.location) BETWEEN -85.05112878 AND 85.05112878
+    ON CONFLICT (cx, cy) DO NOTHING;;
+
+    PERFORM pg_notify('tile_dirty', '');;
```
jake-low: I'm not sure using `pg_notify` here is worth the complexity. As written, this will generate a notification each time a task is modified. I can imagine some workflows (e.g. bulk creation of many large challenges with tens of thousands of tasks each) where this could become a perf problem.

Removing `PERFORM pg_notify()` everywhere would just mean that tiles aren't regenerated immediately after task updates, but they're still rebuilt on a timer every 30s in `TileDirtyBuilder`.

If we remove the notifications and just rely on a timer to rebuild dirty tile cells, it would require some refactoring (`TileDirtyBuilder` wouldn't need a notification listener or its own dedicated DB connection, and would probably need to be converted into a scheduled job instead). But I think it would simplify the architecture with very little negative impact on UX.

app/org/maproulette/models/dal/TaskDAL.scala
```diff
@@ -815,6 +815,19 @@ class TaskDAL @Inject() (
...
+    // websocket event below) hits fresh bytes instead of racing the background
+    // listener. The drain recomputes the affected leaf cells and rolls the
+    // change up to z=0; the bounded limit keeps bulk imports off the critical
+    // path. z=12 itself is served live, so it needs no rebuild here.
+    try {
+      this.serviceManager.tileAggregate.rebuildRecentDirtyCells(limit = 128)
+    } catch {
+      case e: Exception =>
+        logger.warn(s"Post-commit dirty-cell drain failed: ${e.getMessage}")
+    }
```
jake-low: This should be wrapped in `Future {}` I think - rebuilding the dirty cells requires acquiring a global lock. As written, this will block the HTTP response until the lock is available and until all dirty cells have been rebuilt, which isn't necessary from the caller's perspective.

app/org/maproulette/framework/repository/TileAggregateRepository.scala
```diff
@@ -0,0 +1,390 @@
...
+  *     unfiltered requests.
+  *   - Zoom 12: served live from `tasks` as individual / overlap-deduped
+  *     markers (one feature per distinct ground location). The frontend
+  *     overzooms this for z = 13+.
+  *
+  * All four code paths (this repository's live queries, plus `rebuild_leaf_cell`
+  * and `rebuild_all_tile_cells` in evolution 107) share one eligibility filter:
+  * a task is available work when it has a valid location, `status IN (0,3,6)`,
+  * is not archived, and its challenge/project are enabled and not deleted or
+  * archived. Keep them in sync.
```
jake-low: Is `enabled` equivalent to what's called "discoverable" in the maproulette.org UI? If not, I think we should add that as a filter condition too - the explore page map shouldn't show tasks that belong to challenges/projects which are hidden.

app/org/maproulette/framework/service/NominatimService.scala
```diff
@@ -0,0 +1,226 @@
...
+  *
+  * Polygons are resolved by stable OSM identifiers (osm_type + osm_id), not by
+  * Nominatim's internal place_id. Nominatim's public instance does not guarantee
+  * place_id stability across endpoints — the same place_id returned by /search
+  * can resolve to an unrelated OSM object via /details.php (observed in
+  * practice: a city search returning a place_id that /details.php resolves to
+  * a small waterway). Using /lookup with osm_ids avoids that ambiguity.
+  */
+@Singleton
+class NominatimService @Inject() (wsClient: WSClient)(implicit ec: ExecutionContext) {
```
jake-low: I don't think the backend should talk to Nominatim (i.e. we should entirely remove this service from this PR).

This service is currently only used to resolve `(osm_type, osm_id)` pairs to polygons, which is used in three places:
- in `ChallengeDAL.exploreChallenges`, we filter out challenges that don't intersect with this OSM element's polygon. BUT, we're only testing whether the challenge's bbox and the polygon's bbox intersect. The full shape and resolution of the polygon is wasted here
- in `TaskClusterRepository`, we omit tasks that are outside of the provided OSM element's polygon.
- in `TileAggregateService.getMvtTile`, we likewise omit tasks that are outside the polygon

But the maproulette4 frontend already depends on Nominatim: in `LocationSeachFilter.tsx`, it uses Nominatim to look up a user's free-text query for a named place (e.g. "France"), and in response gets back both the OSM element type+ID and the polygon as GeoJSON.

So I think we should remove the Nominatim dependency from the backend. Here's what I think we should do instead:
- For filtering challenges, the frontend should compute the bbox of the returned GeoJSON polygon (e.g. using Turf.js) and then pass just the bbox to the `exploreChallenges` endpoint.
- We should either drop the ability to filter clusters, or require the client to pass a bbox or GeoJSON polygon directly (rather than an OSM ID) in order to filter them.
- We should drop the ability to filter the contents of MVT tiles. The tile endpoint should not accept an `?osm_id=` query parameter _or_ a `?polygon=` parameter, because doing either means that tiles cannot efficiently be cached at the HTTP layer. There are millions of possible OSM IDs that might be used to filter, and effectively infinite polygons, so each user that applied a spatial filter to the tile data would effectively be creating a new unique tileset (containing millions of potential tiles) that needs to be generated by the backend and then be persisted into the HTTP cache (even though it's unlikely anyone else would later request the exact same data). MVT tiles are typically not generated dynamically in response to user requests for this reason: they should either be generated statically in advance, or on demand, but either way they should be the same for all users, and any customization should be applied client-side using MapLibre styles/filters/etc. In our case, I think the existing UX (show tasks even if they are outside the filter area) is fine, since we highlight the filter area itself on the map.

app/org/maproulette/framework/service/NominatimService.scala
```diff
@@ -0,0 +1,226 @@
...
+          .map { point =>
+            val coords = point.as[JsArray]
+            s"${coords(0).as[Double]} ${coords(1).as[Double]}"
+          }
+          .mkString(", ")
+        s"($points)"
+      }
+      .mkString(", ")
+    s"POLYGON($rings)"
+  }
```
jake-low: Nit (not relevant since I think we should remove this entire service, but just FYI): JTS's `GeoJSONReader` and `WKTWriter` can do this conversion; no need to handroll it.

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -2090,4 +2488,110 @@ class ChallengeDAL @Inject() (
...
+      // Filter by keywords if provided
+      keywords match {
+        case Some(kws) if kws.trim.nonEmpty =>
+          val keywordList = kws.split(",").map(_.trim.toLowerCase).filter(_.nonEmpty)
+          if (keywordList.nonEmpty) {
+            val keywordConditions = keywordList.map(kw => s"LOWER(t.name) = '$kw'").mkString(" OR ")
+            query += s" AND ($keywordConditions)"
+          }
+        case _ =>
+      }
```
jake-low: This makes me nervous since keywords is a user-supplied string. Can't this be used as a SQL injection vector?

app/org/maproulette/framework/repository/CommentRepository.scala
```diff
@@ -102,6 +102,46 @@ class CommentRepository @Inject() (override val db: Database) extends Repository
...
+    * @param searchTerm The term to search within the comments
+    * @param limit The maximum number of comments to return
+    * @param page The page number for pagination
+    * @return A list of matching Comments
+    */
+  def searchComments(
+      searchTerm: String,
+      limit: Int = 25,
+      page: Int = 0
+  )(implicit c: Option[Connection] = None): List[Comment] = {
```
jake-low, line 123: Two concerns with this function:
- It allows unauthenticated search over all task comments; is this really what we want? I think maybe it should require auth, and filter to either comments the user wrote, or comments on challenges the user owns.
- I don't think there's a trigram index on `task_comments.comment`, so the `ILIKE {searchterm}` will do a full scan over the entire table, which is going to be pretty slow. We're already adding pg_trgm for task name search; we should probably build an index over this column too if we're going to support full text search here.

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -1015,14 +1270,78 @@ class ChallengeDAL @Inject() (
...
+              ST_Y(tasks.location) as lat,
+              ST_X(tasks.location) as lng,
+              tasks.status,
+              tasks.priority,
+              tasks.bundle_id,
+              l.user_id as locked_by,
+              ST_ClusterDBSCAN(tasks.location, eps := 0.000001, minpoints := 1) OVER () as cluster_id
+            FROM tasks
+            LEFT JOIN locked l ON l.item_id = tasks.id AND l.item_type = 2
+            WHERE tasks.parent_id = $id"""
```
jake-low: This query doesn't accept pagination parameters and doesn't have a default LIMIT (it returns all tasks in the challenge, even for huge challenges). Is that deliberate? The existing `GET /challenge/:id/tasks` parameter does support pagination.

app/org/maproulette/framework/controller/TaskController.scala
```diff
@@ -128,6 +128,72 @@ class TaskController @Inject() (
...
+      val location = if (bounds.nonEmpty) {
+        try {
+          val coords = bounds.split(",").map(_.trim.toDouble)
+          if (coords.length == 4) {
+            Some(SearchLocation(coords(0), coords(1), coords(2), coords(3)))
+          } else {
+            None
+          }
+        } catch {
+          case _: Exception => None
```
jake-low: Let's at least add a log statement here if we're going to swallow errors.

EDIT: actually, this would trigger if the client supplied an invalid `bbox` string in the request, right? I think that should trigger a 400 error, not silently ignore the bbox parameter.

app/org/maproulette/controllers/api/TaskController.scala
```diff
@@ -825,4 +848,130 @@ class TaskController @Inject() (
...
+      taskIds.distinct.flatMap(this.dal.retrieveById(_)).partition { task =>
+        try {
+          dalManager.challenge.retrieveById(task.parent) match {
+            case Some(challenge) =>
+              permission.hasWriteAccess(ProjectType(), user)(challenge.general.parent)
+              true
+            case None => false
+          }
+        } catch {
+          case _: Exception => false
```
jake-low: Add a log statement here or (preferably) raise an exception back to the caller.

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -946,40 +1140,101 @@ class ChallengeDAL @Inject() (
...
+          // with its old priority.
+          currentTasks = listChildren(DEFAULT_NUM_CHILDREN_LIST, pointer)
+
+          // Evaluate per task with a guard: a single malformed rule/bound shouldn't
+          // abort the whole recompute and leave the batch half-written. On failure
+          // we preserve the task's current priority so it stays internally consistent.
+          val evaluated: List[(Task, Int)] = currentTasks.map { task =>
+            val p =
+              try task.getTaskPriority(challenge)
+              catch { case _: Throwable => task.priority }
```
jake-low: I think catching `Throwable` is too wide here (pretty sure that includes things like OOM errors). Probably want to catch `Exception` instead (if not something more specific).

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -946,40 +1140,101 @@ class ChallengeDAL @Inject() (
...
+      val draftChallenge           = persisted.copy(priority = draft)
+      val result                   = scala.collection.mutable.LongMap[Int]()
+      var pointer                  = 0
+      var currentTasks: List[Task] = List.empty
+      do {
+        currentTasks = listChildren(DEFAULT_NUM_CHILDREN_LIST, pointer)
+        currentTasks.foreach { task =>
+          val p =
+            try task.getTaskPriority(draftChallenge)
+            catch { case _: Throwable => task.priority }
```
jake-low: Ditto above (Throwable is too generic)

app/org/maproulette/models/dal/ChallengeDAL.scala
```diff
@@ -1056,6 +1375,59 @@ class ChallengeDAL @Inject() (
...
 
+  def search(
+      search: String,
+      limit: Int = 25
+  )(implicit c: Option[Connection] = None): List[Challenge] = {
+    this.withMRConnection { implicit c =>
+      val isNumeric     = search.matches("^\\d+$")
+      val searchLong    = if (isNumeric) Some(search.toLong) else None
+      val searchPattern = if (search.nonEmpty) s"%${search.replace("'", "''")}%" else "%"
+      val searchLower   = if (search.nonEmpty) search.toLowerCase.replace("'", "''") else ""
```
jake-low: - `searchLower` is unused
- `searchPattern` is bound (not interpolated directly) into the SQL query, so it doesn't need quote-escaping (drop `.replace("'", "''")`)

app/org/maproulette/framework/controller/ProjectController.scala
```diff
@@ -211,6 +211,27 @@ class ProjectController @Inject() (
...
+    * @return A single project matching the search criteria, or empty list if not found
+    */
+  def search(
+      search: String,
+      onlyEnabled: Boolean = false
+  ): Action[AnyContent] = Action.async { implicit request =>
+    this.sessionManager.userAwareRequest { implicit user =>
+      val results = this.projectService.search(
+        search
+      )
```
jake-low: The controller accepts `onlyEnabled` but this parameter is unused. Also, the service accepts a `limit` param but the endpoint/controller do not. Both should probably be plumbed all the way through.

app/org/maproulette/models/utils/ChallengeFormatters.scala
```diff
@@ -108,6 +110,23 @@ trait ChallengeWrites extends DefaultWrites {
...
   )(unlift(Challenge.unapply))
+
+  // Frontend Challenge type expects a completionMetrics object; the
+  // Challenge case class only carries tasksRemaining/completionPercentage,
+  // so per-status counters fall back to CompletionMetrics defaults (zero).
+  implicit val challengeWrites: Writes[Challenge] = Writes { c =>
+    val base      = challengeFieldsWrites.writes(c).as[JsObject]
+    val remaining = c.tasksRemaining.getOrElse(0)
+    val total = c.completionPercentage match {
+      case Some(pct) if pct > 0 && pct < 100 =>
```
jake-low: I think you need a special case for `pct == 100` (as is, it'll return `total: 0`). Don't just change `pct < 100` to `pct <= 100` though or you'll divide by 0

jake-low (changes requested):
Finished a first pass review. Left lots of comments, but my ~~two~~ three (I can count I swear) main areas of concern are the breaking changes to the JSON response schema of `GET /challenge/:id`, use of `pg_notify` for triggering rebuilds of dirty tile cells (vs a simple timer task), and the dependency on Nominatim in the backend (vs confining that dependency to just the frontend code).

app/org/maproulette/controllers/api/ChallengeController.scala
CollinBeczak: So this is a little confusing... i tried in swagger but the endpoint outputed the type you pointed out.

Though in maproulette.org, the endpoint:
GET https://maproulette.org/api/v2/challenge/42238
returns
```
{
    "id": 42238,
    "name": "Toronto - Missing Crossing Way",
    "created": "2023-11-12T08:51:29.649Z",
    "modified": "2026-05-28T23:16:46.759Z",
    "description": "This challenge focuses on identifying\ncrossing nodes that are not connected to a crossing way. A crossing *node* is\na specific type of node that indicates where pedestrian or bicycle crossings\nintersect with roads. A crossing *way*, on the other hand, represents the\nphysical crossing itself, such as a lined or zebra crossing.\n\nTypically, a crossing node should connect to both a highway and a footway\nor cycleway. However, we have found instances where crossing nodes are either\nincorrectly tagged or not connected to any crossing highway. Each task flags\nan instance of a crossing node that is not connected to a crossing way.\n\nWe ask that mappers working on these tasks use satellite imagery to identify\nthe crossing node and its connection to the highway. If the crossing node is\nis mistagged, or misplaced then please resolve the issue and make sure that\nthe crossing node is located at the intersection of a crossing way and a\nhighway.\n",
    "deleted": false,
    "isGlobal": false,
    "requireConfirmation": false,
    "requireRejectReason": false,
    "infoLink": "",
    "owner": 9848585,
    "parent": 56810,
    "instruction": "It appears that {{osmIdentifier}} is\nnot tagged correctly. Please ensure that the node is located at the\nintersection of a crossing way and a highway. If either a highway or a\ncrossing way is missing, it should be added accordingly. If two ways intersect\nat this node, verify that both are appropriately tagged to distinguish one as\na highway and the other as a crossing.\n\nPlease adjust or add the following tags on the node:\n- crossing=*: This tag identifies the type of crossing, such as an\nuncontrolled crossing, an unmarked crossing, or a crossing with traffic\nsignals.\n- crossing:markings=*: This tag indicates the type of markings present on\nthe crossing, including options like zebra stripes, painted lines, or surface\nchange.\n\nFurthermore, it is crucial to ensure that the tagging of the crossing way is\naccurate. All tags on the crossing node should be consistent with those on\nthe crossing way to maintain clarity and accuracy in the mapping data.\n\nIf you are looking for more information on crossings, be sure to check out these\nwiki pages:\n[crossings wiki page](https://wiki.openstreetmap.org/wiki/Crossings)\n[key:crossing wiki page](https://wiki.openstreetmap.org/wiki/Key:crossing)\n[key:crossing:markings](https://wiki.openstreetmap.org/wiki/Key:crossing:markings)\n",
    "difficulty": 2,
    "blurb": "",
    "enabled": true,
    "featured": false,
    "cooperativeType": 0,
    "popularity": 1780010093,
    "checkinComment": "Added crossing way tags. #MissingCrossingWay #maproulette",
    "checkinSource": "",
    "requiresLocal": false,
    "overpassQL": "",
    "remoteGeoJson": "https://daylight-map-distribution.s3.us-west-1.amazonaws.com/mr/osm_checks/Toronto_MissingCrossingWay.json",
    "overpassTargetType": "",
    "defaultPriority": 0,
    "highPriorityRule": {
        "condition": "AND",
        "rules": [
            {
                "value": "priority.0",
                "type": "integer",
                "operator": "=="
            }
        ]
    },
    "mediumPriorityRule": {
        "condition": "AND",
        "rules": [
            {
                "value": "priority.1",
                "type": "integer",
                "operator": "=="
            }
        ]
    },
    "lowPriorityRule": {
        "condition": "AND",
        "rules": [
            {
                "value": "priority.2",
                "type": "integer",
                "operator": "=="
            }
        ]
    },
    "highPriorityBounds": [],
    "mediumPriorityBounds": [],
    "lowPriorityBounds": [],
    "defaultZoom": 13,
    "minZoom": 1,
    "maxZoom": 19,
    "updateTasks": false,
    "limitTags": false,
    "limitReviewTags": false,
    "isArchived": false,
    "reviewSetting": 0,
    "defaultBasemap": -1,
    "defaultBasemapId": "",
    "customBasemap": "",
    "exportableProperties": "",
    "osmIdProperty": "identifier",
    "taskBundleIdProperty": "",
    "taskWidgetLayout": "",
    "status": 3,
    "statusMessage": "",
    "lastTaskRefresh": "2026-04-05T09:13:14.253Z",
    "dataOriginDate": "2026-04-05T00:00:00.000Z",
    "location": {
        "type": "Point",
        "coordinates": [
            -79.518646478,
            43.687672796
        ]
    },
    "bounding": {
        "type": "Polygon",
        "coordinates": [
            [
                [
                    -79.853443976,
                    43.493652593
                ],
                [
                    -79.853443976,
                    43.839958405
                ],
                [
                    -79.123624228,
                    43.839958405
                ],
                [
                    -79.123624228,
                    43.493652593
                ],
                [
                    -79.853443976,
                    43.493652593
                ]
            ]
        ]
    },
    "completionPercentage": 61,
    "tasksRemaining": 3009,
    "tags": [
        "missing_crossing_way"
    ]
}
```

Which aligns with the new type i made. Thats why i did it is because to me it looks like the backend doesnt actually return what it says it returns. But at the same time swagger does output what it says it was going to so idk. Maybe its just swagger filling in defaults if the data doesnt match?

jake-low: Looks like you're right; the Swagger docs incorrectly show a nested JSON schema for the challenge endpoint, but the actual current behavior appears to use a flattened schema like the one represented by `BaseChallenge`. So AFAICT this isn't actually a breaking change.

app/org/maproulette/jobs/TileDirtyListener.scala
CollinBeczak: Done — converted in commit https://github.com/maproulette/maproulette-backend/pull/1237/changes/fd4ac0334c4f4f28ddf8b9f6d53956c160fadf25.

- `PERFORM pg_notify('tile_dirty', '')` removed from all three triggers in `107.sql`.
- `TileDirtyListener` deleted along with its dedicated `DriverManager` connection and daemon thread.
- The drain now lives in a scheduled `SchedulerActor` job (`rebuildDirtyTileCells`) — `app/org/maproulette/jobs/SchedulerActor.scala:112`, registered in `app/org/maproulette/jobs/Scheduler.scala:119`. Default interval is 30s (`maproulette.scheduler.rebuildDirtyTileCells.interval`; `test.conf` sets it to `""` to disable in tests).

app/org/maproulette/models/dal/TaskDAL.scala
CollinBeczak: I will just remove this for now and rely on the scheduled action.

app/org/maproulette/framework/repository/TileAggregateRepository.scala
CollinBeczak: yes, discoverable is set by the enabled value.

app/org/maproulette/models/dal/ChallengeDAL.scala
CollinBeczak: k, replaced with Exception

app/org/maproulette/models/dal/ChallengeDAL.scala
CollinBeczak: replaced with exception

app/org/maproulette/controllers/api/TaskController.scala
CollinBeczak: Went with raising an exception. If any task is missing or the caller lacks write access, resolveTasksWithWriteAccess now throws and the whole bulk request fails with 404/403 instead of silently skipping.

app/org/maproulette/framework/controller/TaskController.scala
CollinBeczak: A malformed bounds (or challengeIds) now returns 400 with an explanatory message instead of being silently dropped.

app/org/maproulette/models/dal/ChallengeDAL.scala
CollinBeczak: This was deliberate, though maybe we should add a limit of 50,000 just in case. The goal is to just fetch all the tasks in the challenge once and do pagination, filtering client side, and updates via websockets.

CollinBeczak: i added a InvalidException error if the challenge has more than 50,000 tasks.

app/org/maproulette/framework/repository/CommentRepository.scala
CollinBeczak: I added the indexes and the user filter

app/org/maproulette/models/dal/ChallengeDAL.scala
CollinBeczak: Good catch — fixed. Keywords are now bound as named parameters and matched via LOWER(t.name) IN ({kw0}, {kw1}, …), so no user input ends up in the SQL string. Only the placeholder count is interpolated, derived from the list size. Same accepted inputs and matching semantics as before for legitimate keywords; injection strings just return no rows now instead of executing.

jake-low (approved):
LGTM!

Edit: looks like there are merge conflicts that need to be resolved; let me know if you want a hand with that and I can take a look at them

CollinBeczak:
@jake-low, i dont see any merge conflicts

jake-low:
This is what I see in the merge UI:

image

EDIT: Oh, I see, it's only showing this for the rebase option but squashing shows no conflicts. I'm fine with squashing this; merge whenever you're ready (we can wait until tomorrow when you have more time to cut a release and deploy this)

CollinBeczak:
Sounds good, it can wait until tomorrow.

PR #1237, opened 2026-05-29
Resolve review comments pr 1236

CollinBeczak:
# Address first-pass review feedback

This PR works through the review comments from the first pass. The changes are
split into focused commits (one per issue, with a few tightly-coupled or
duplicate comments combined). Below is an overview of the three main areas of
concern, then the full commit-to-comment map.

## The three big-ticket items

**1. Nominatim in the backend → removed.** `NominatimService` is gone. The
frontend already resolves place names to geometry via Nominatim, so the backend
no longer needs to. `exploreChallenges` now filters by a client-supplied
bounding box, the task cluster/marker queries drop server-side spatial
filtering, and the MVT tile endpoint no longer takes `osm_id`/`polygon` — so
tiles stay HTTP-cacheable. (The matching frontend change — sending `bounds`
instead of osm ids — lives in the maproulette4 repo and isn't part of this PR.)

**2. `pg_notify` for tile rebuilds → scheduled job.** All `pg_notify` calls are
removed from the tile triggers, `TileDirtyListener` (and its dedicated LISTEN
connection) is deleted, and the synchronous post-commit drain is removed from
`TaskDAL`. A `SchedulerActor` job (`rebuildDirtyTileCells`, default 30s) now
drains the dirty-cell queue and logs when it can't keep up. This simplifies the
architecture at the cost of up to ~30s tile-refresh latency, which is fine for
the explore map.

**3. Breaking JSON shape of `GET /challenge/:id`.** This one is a deliberate
judgment call worth flagging: **the flat (`BaseChallenge`) shape is kept**,
because maproulette4 is built to consume the challenge response flat
(`challenge.owner`, `challenge.instruction`, …) rather than nested
(`challenge.general.owner`). Reverting to the nested shape would break the new
frontend. What I *did* fix is the related accuracy concern: the legacy
`Challenge` endpoints (`list`/`find`/`extendedFind`) now read the real
`completion_metrics` JSONB column too, so they return the same accurate
per-status metrics as `GET /challenge/:id` instead of synthesized/zeroed ones.
Net result: the "wrong stats" inconsistency is resolved; the flat-vs-nested
shape difference between `GET /:id` and the list endpoints remains a conscious
choice (maproulette3 is being phased out in favor of maproulette4).

## Commit → review-comment map

Comments are numbered in the order they appear in the first-pass review (top to
bottom). Note that **comment 1** (the breaking flat shape on `GET /challenge/:id`)
has no commit — see item 3 above; the flat shape is kept intentionally.

| Commit | Review comment(s) | What it does |
|--------|-------------------|--------------|
| `33b082e1` | comment:2 (wrong synthesized metrics), comment:21 (`pct == 100` returns 0) | Read accurate `completion_metrics` from JSONB; `completionMetrics` replaces top-level `tasksRemaining` |
| `a6aaf7c4` | comment:3 (misplaced doc comment + affected-rows signaling) | Move the `updateTaskPriorities` doc onto the right function; clarify failures raise and `(0,0,0)` is legitimate |
| `31744d16` | comment:18 (`catch Throwable`, two spots) | Narrow priority-eval guards to `Exception` |
| `81dc1238` | comment:15 (unpaginated `taskMarkers`) | Add `limit`/`page`; cluster over just the requested page |
| `b99786df` | comment:19 (dead `searchLower`/escaping), comment:20 (unused `onlyEnabled`, unplumbed `limit`) | Plumb `onlyEnabled`/`limit` through challenge & project search; drop dead vars |
| `01d8df33` | comment:11 (backend Nominatim), comment:12 (`polygonToWKT` nit), comment:13 (keyword SQL injection) | Remove `NominatimService`; bbox/geometry filtering; parameterize all keyword filters |
| `5fc51ea1` | comment:4 (buffered bounding box) | Use the exact `ST_Extent` (no buffer) |
| `fd4ac033` | comment:7 (log when drain is busy), comment:8 (`pg_notify` complexity), comment:9 (wrap drain in `Future`) | Scheduled `rebuildDirtyTileCells` job replaces the listener/notify/post-commit drain |
| `b2b0945b` | comment:10 (is `enabled` == discoverable?) | Document that `enabled` is the discoverable flag (already required on all eligibility paths) |
| `b5f2ade3` | comment:14 (unauthenticated comment search + missing trigram index) | Require auth, scope to author/owned challenges, add pg_trgm indexes (evolution 113) |
| `b608927b` | comment:16 (swallowed bbox error), comment:17 (swallowed write-access error) | Return 400 on bad bbox; catch only `IllegalAccessException` and log the rest |
| `93780f3a` | comment:5 (105.sql), comment:6 (106.sql) | Revert 105/106 to their original content |

## Notes for reviewers / deploy

- **Migrations:** 105 and 106 are restored to their original content, so the
  checksum-mismatch workaround (manually editing `play_evolutions`) that the
  earlier version of this PR needed is **no longer required**. One new migration
  is added — **113** (pg_trgm GIN indexes on `task_comments.comment` and
  `challenge_comments.comment`).
- **API shape:** `GET /challenge/:id` stays flat (`BaseChallenge`);
  `list`/`find`/`extendedFind`/`PUT` keep the nested `Challenge` model. If we'd
  rather make the whole challenge API internally consistent, the follow-up is to
  flatten the list endpoints too — happy to do that in a separate PR.
- **Frontend:** the maproulette4 side (send `bounds`/geometry instead of
  `osm_type`/`osm_id`, drop the tile spatial filter) is a coordinated change in
  that repo and should land before/with deploy.


CollinBeczak:
Defects extended to their sibling locations:

Commit b5f2ade3 (comment:14): Reviewer flagged unauthenticated/full-scan search only on CommentRepository (task comments). I applied the same auth-required + scoping + trigram index to ChallengeCommentRepository (challenge comments) too — same code pattern, same vulnerability.
Commit b99786df (comment:20): Reviewer flagged unused onlyEnabled / unplumbed limit only on ProjectController.search. I also plumbed it through ChallengeController.search (identical defect).
Commit b99786df (comment:19): Reviewer flagged unused searchLower / redundant escaping only on ChallengeDAL.search. I also cleaned it up in ProjectService.search (identical defect).
Commit 01d8df33 (comment:13): Reviewer flagged keyword SQL injection only on ChallengeDAL.exploreChallenges. I also parameterized the keyword filters in all 4 TaskClusterRepository query methods (same pattern, escaped-but-still-interpolated).
Reviewer-endorsed alternative chosen explicitly:

Commit 01d8df33: For the cluster/marker endpoints the reviewer offered two options — "drop the ability to filter clusters" or "require the client to pass bbox/polygon directly." Per your subsequent instruction, I went with drop: those endpoints have no spatial-filter param at all (no osm_type/osm_id, no client polygon).
Small structural addition supporting comment:13:

TaskClusterRepository got three small private helpers (parseKeywords, keywordNamedParameters, keywordInClause) so the four query methods share the parameterized-keyword pattern instead of each open-coding it. Pure refactor in service of the SQL-injection fix.
Things that look extra but aren't:

The new evolution 113.sql (pg_trgm indexes) — explicitly asked for in comment:14.
The new rebuildDirtyTileCells scheduler job + Config key + application.conf / test.conf entries — required by comment:8.
The TestSpec.scala change removing the NominatimService mock — required by comment:11.
All the route-doc edits (challenge.api, task.api, project.api) — required by the parameter signature changes those comments mandated.

PR #1239, opened 2026-06-01
Fix Swagger API docs to correctly describe types of embedded JSON values

jake-low:
The OpenAPI spec published at `/assets/swagger.json` claimed several fields were plain strings (containing stringified JSON values) when in fact they are always serialized directly as embedded JSON. This caused the generated TS types in `openApiTypes.ts` in the maproulette4 codebase to be wrong for these fields, which required defensive workarounds like `typeof x === 'string' ? JSON.parse(x) : x` in several places.

This PR modifies the backend models and serialization code in the controllers to correct this. There were two different issues that caused this:
1. `Task.geometries`, `Task.location`, `Task.cooperativeWork`, and `User.properties` were declared as `String` / `Option[String]` on their case classes, even though the column contents are JSON. Each model's custom Play JSON `Format` called `Json.parse(...)` and re-inserted the value as a parsed object before emitting. The Swagger types are generated by reflection, which can only see the type annotations and not the custom `Format` behavior which differs from the declared types.
2. `conf/swagger-custom-mappings.yml` contained a global mapping from `play.api.libs.json.JsValue` to `string`, which is wrong. JsValue can contain any JSON type (object, array, string, number, or null).

This PR addressed these problems.
- I've refactored the code to parse JSONB values from Postgres at read time, rather than read them as strings and then parse them when serializing JSON payloads for HTTP responses. This means the models can have their actual types (JsValue, JsObject, etc) correctly annotated instead of using custom Format code, so the Swagger generator sees accurate type information.
- I've changed this to map JsValue to the "top" type (TypeScript's `any` or `unknown`).
- I've also refactored various places where JsValue was used in response types to use a more specific type (like JsObject or JsArray) instead, and I added a migration to clean up some string-typed sentinel values in the database in order to make this new code simpler.

### DIFF
--- a/app/org/maproulette/controllers/api/TaskController.scala
+++ b/app/org/maproulette/controllers/api/TaskController.scala
@@ -133,46 +133,21 @@ class TaskController @Inject() (
     this.updateGeometryData(super.updateUpdateBody(body, user))
 
   private def updateGeometryData(body: JsValue): JsValue = {
-    val updatedBody = (body \ "geometries").asOpt[String] match {
-      case Some(value) =>
-        // if it is a string, then it is either GeoJSON or a WKB
-        // just check to see if { is the first character and then we can assume it is GeoJSON
-        if (value.charAt(0) != '{') {
-          // TODO:
-          body
-        } else {
-          // just return the body because it handles this case correctly
-          body
-        }
-      case None =>
-        // if it maps to None then it simply could be that it is a JSON object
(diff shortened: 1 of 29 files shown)


CollinBeczak:
LGTM, code looks good and tested against local mr3 and mr4 and I didnt see any issues. Had to update some types in mr4, but otherwise is good.

CollinBeczak (approved):

PR #1240, opened 2026-06-02
Drop unused task_geometries table and lazy migration in read path

jake-low:
This table was once used to store task geometries. Commit b95bfe2d moved this data into the tasks table instead, but the old table was kept and data was migrated lazily on read by Scala code.

Even though all data has been migrated and the task_geometries is completely unused now, the table and corresponding code was never removed.

This commit also adds a NOT NULL constraint to task.geometries. In the production database, this constraint is already true, so adding it is an easy way to prevent invalid tasks with no geometry from being created in the future.

Note: this PR is based on top of the `add-support-for-new-features` branch, in order to avoid merge conflicts (since both branches add new evolutions). But we don't need to merge it together with that branch. Once the upstream branch merges, GitHub will automatically change this PR's target to `main`.


CollinBeczak:
LGTM

PR #1241, opened 2026-06-02
Validate task GeoJSONs and ensure task.location and task.geom are non-NULL

jake-low:
Followup on top of #1240, which ensures that task GeoJSONs are not null.

By applying additional validations at insert time, we can ensure that:
- task.geojson is not only non-null, but is actually a well formed GeoJSON FeatureCollection
- task.geom and task.location (which are computed from that GeoJSON) are also non-null

This PR does both. It adds some basic validation on the task GeoJSON (to make sure it's a valid and non-empty FeatureCollection) at insert/update time, and it adds an evolution to apply NOT NULL constraints to the geom and location columns.

Before deploying this, I will need to do some minor manual database maintenance, because there are about 1,000 tasks (mostly very old) that have null locations or geoms. I plan to delete these tasks, since they are not actionable without this info (a few have comments attached to them by mappers, and the comments are universally by confused users wondering why MapRoulette sent them to [Null Island](https://en.wikipedia.org/wiki/Null_Island) and there isn't anything available to fix).

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -322,6 +322,7 @@ class TaskDAL @Inject() (
       user: User
   )(implicit id: Long, c: Option[Connection] = None): Option[Task] = {
     this.permission.hasObjectWriteAccess(element, user)
+    validateGeoJson(element.geometries)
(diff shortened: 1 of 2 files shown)


CollinBeczak:
LGTM

PR #1246, opened 2026-06-04
Remove updateChallengeCompletionMetrics scheduled job

jake-low:
#1236 added a new `completion_metrics` JSONB column which is kept up to date by triggers, but it did not remove the scheduled job that maintains the old `completion_percentage` and `tasks_remaining` counts. Having both at once creates write contention on the challenges and projects tables and makes lots of API operations very slow while the job is running.

This PR does the following:
- drops the `completion_percentage` and `tasks_remaining` columns (and computes those fields from the `completion_metrics` data when building JSON responses)
- removes the `updateCompletionMetricsOfActiveChallenges` job which maintained those columns
- changes when the `update_challenge_completion_metrics` and `update_project_completion_metrics` SQL functions are triggered, so updating a task (or challenge) will only recompute metrics for the parent challenge (or project) if a column was modified that could affect the results.

There are still problems with the present approach which we should fix in a future PR:
- The metrics calculation SQL functions are triggered per row, not per statement. This means that a bulk operation (like inserting 10,000 new tasks) calls the function 10,000 times, when calling it once at the end would be sufficient.
- The code paths for creating a challenge or refreshing tasks build task models one at a time and then `INSERT` or `UPDATE` them in individual queries. This is an antipattern on its own (refreshing a challenge with 50k tasks means 50k DB roundtrips), but it also means that even if we switch to per-statement triggers as described above, we still won't see any benefit until these code paths are refactored to insert/update tasks in batches.

### DIFF
--- a/app/org/maproulette/Config.scala
+++ b/app/org/maproulette/Config.scala
@@ -336,8 +336,6 @@ object Config {
     s"$SUB_GROUP_SCHEDULER.archiveChallenges.startTime"
   val KEY_SCHEDULER_ARCHIVE_CHALLENGES_STALE_TIME_IN_MONTHS =
     s"$SUB_GROUP_SCHEDULER.archiveChallenges.staleTimeInMonths"
-  val KEY_SCHEDULER_UPDATE_CHALLENGE_COMPLETION_INTERVAL =
-    s"$SUB_GROUP_SCHEDULER.updateChallengeCompletionMetrics.interval"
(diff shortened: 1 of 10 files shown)


CollinBeczak:
LGTM, i tested locally and things work well in both frontends. Will keep the problems you noted in mind.

CollinBeczak (approved):

PR #1247, opened 2026-06-26
Add tests for service info, user auth, task geometry pruning, and overpass task building

ljdelight:
Create a few test cases to get familiar with the framework and a more complex test to start validating overpass (via a mocked request).

- Service info endpoint: verify build and runtime info is returned
- User whoami: cover valid apiKey, session, and rejection auth paths
- Task geometry pruning: regression test for ClassCastException on Json.arr-backed arrays (#1243)
- Overpass challenge building: success, wrong content-type, and 504 busy-server failure paths

### DIFF
--- a/app/org/maproulette/models/dal/TaskDAL.scala
+++ b/app/org/maproulette/models/dal/TaskDAL.scala
@@ -496,16 +496,8 @@ class TaskDAL @Inject() (
         cooperativeWorkJson = Some(workMatch.head.toString())
       }
 
-      val attachments   = (geometries \ "attachments").toOption
-      val mrTransformer = (__ \ "properties" \ "maproulette").json.prune
-      val extractedGeometries = JsArray(
-        (geometries \ "features")
-          .as[JsArray]
-          .value
-          .collect {
-            case value: JsObject => value.transform(mrTransformer).getOrElse(value)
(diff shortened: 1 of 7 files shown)


ljdelight:
Fixed the inconsistent issue with a test, this is ready.

CollinBeczak (approved):

jake-low (approved):
I reviewed the code changes and skimmed the new test code; LGTM.

> start validating overpass (via a mocked request)

Just to double check: this means we aren't sending any requests to overpass-api.de when the test suite runs, right? (I don't want to generate any unnecessary load for that service by pinging it every time our CI runs)

ljdelight:
> I reviewed the code changes and skimmed the new test code; LGTM.
> 
> > start validating overpass (via a mocked request)
> 
> Just to double check: this means we aren't sending any requests to overpass-api.de when the test suite runs, right? (I don't want to generate any unnecessary load for that service by pinging it every time our CI runs)

That's exactly right, the request doesn't visit the real server. The test is safe from overpass going down for updates or changes in its response structure (say, more element nodes are in a response).

The mocked response is hard-coded in the test file and what we can do is copy this style to do additional validation around any pain points with overpass.

PR #1248, opened 2026-06-29
Respect client-requested redirect_uri in OAuth flow

CollinBeczak:
The OAuth redirect_uri was hardcoded to config.getMRFrontend in both the authorize request and the token exchange, so a login started on any frontend (e.g. beta.maproulette.org) sharing this backend would be redirected to the backend's single configured frontend after authenticating with OSM.

Both /auth/authenticate and /auth/callback now accept a redirectUri parameter and fall back to the request's Origin header, then the configured frontend. OSM validates redirect_uri against the OAuth app's registered URIs, so no local allowlist is needed.

### DIFF
--- a/app/controllers/AuthController.scala
+++ b/app/controllers/AuthController.scala
@@ -48,78 +48,93 @@ class AuthController @Inject() (
   val logger: Logger = LoggerFactory.getLogger(classOf[AuthController])
   import scala.concurrent.ExecutionContext.Implicits.global
 
+  /**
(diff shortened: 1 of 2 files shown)


jake-low (approved):
LGTM, thanks for implementing this!

This has an additional benefit which isn't mentioned in the PR description: with this change, it'll be possible to use OAuth login when running the frontend locally (i.e. during development) while it's configured to talk to the production backend. Currently this is possible only by setting `REACT_APP_SERVER_API_KEY` (in maproulette3) or `VITE_SERVER_API_KEY` (in maproulette4) to your user account's API key before starting the frontend.

PR #1250, opened 2026-06-29
Add task lock expiry reminder emails

CollinBeczak:
Emails users with an opt-in subscription shortly before one of their locked tasks is about to be auto-unlocked by the cleanLocks job. The reminder window is configurable (default 10 minutes) and the scheduled job is itself opt-in: omit `osm.scheduler.notifications.taskLockExpiryReminder.interval` to leave it unscheduled.

- Adds NOTIFICATION_TYPE_TASK_UNLOCK_WARNING (16) and a `taskUnlockWarning` field on NotificationSubscriptions, defaulting to NOTIFICATION_EMAIL_NONE.
- Adds the `sendTaskLockExpiryReminders` SchedulerActor job, which uses an atomic UPDATE...RETURNING on the `locked` table so overlapping runs cannot double-send.
- Evolution 119 adds `reminder_sent_at` to `locked` and `task_unlock_warning` to `user_notification_subscriptions`.


conf/evolutions/default/119.sql
```diff
@@ -0,0 +1,20 @@
+# --- MapRoulette Scheme
+
+# --- !Ups
+
+-- Track whether an unlock warning email has been sent for a given lock so
+-- the sendTaskLockExpiryReminders job does not email the same user about
+-- the same lock more than once.
+ALTER TABLE IF EXISTS locked
+  ADD COLUMN reminder_sent_at timestamp without time zone DEFAULT NULL;;
```
ljdelight: Why would you remove the timezone? That will break things, the zone should always be tracked so that timings can be correct based on a user's zone (vs the server's configured timezone, which is not reliable).

edit: i know it's a draft, giving you a heads up on that one. also you'd wanna do a 'add column if not exists'.

jake-low:
A task lock expiring and being automatically released is a routine and nondestructive action; is there really a benefit to emailing users when it occurs? I'm worried about the volume of email this has the potential to generate (for us, as a sender). Have we considered other options, like showing a timer in the UI while a task is locked?

CollinBeczak:
Yeah, we can start with that. The goal was to give users some visibility on the lock status when they are in an external editor and not looking at the maproulette ui. But adding a timer in the ui and a modal that allows the user to reset or extend the lock would be a good start.

PR #1252, opened 2026-07-03
Use a separate DB connection pool and execution context for task rebuilding

jake-low:
As discussed in Slack. This introduces a new DB connection pool called `background`, and a new thread pool / execution context called `task-builder-dispatcher` which are used for building (or rebuilding) challenge tasks. The separate DB pool means that the default pool won't be exhausted so requests can still be served, and the separate thread pool means that if a burst of rebuilds are requested at once, only a fixed number will be allowed to progress (equal to the pool size) and the rest will queue up rather than entering a fail-and-retry loop.

This should hopefully help alleviate the issue we've been seeing where API requests time out periodically during times when some user is requesting a whole bunch of challenges be rebuilt in a short burst.

### DIFF
--- a/app/org/maproulette/provider/ChallengeProvider.scala
+++ b/app/org/maproulette/provider/ChallengeProvider.scala
@@ -536,185 +536,181 @@ class ChallengeProvider @Inject() (
                 )
               }
 
-              this.db.withTransaction { implicit c =>
-                var partial          = false
-                val payload          = result.json
-                var targetTypeFailed = false
+              var partial          = false
+              val payload          = result.json
+              var targetTypeFailed = false
 
-                // parse the results. Overpass has its own format and is not geojson
-                val elements = (payload \ "elements").as[List[JsValue]]
-                try {
-                  elements.foreach { element =>
-                    // Verify target type if we are given one.
-                    challenge.creation.overpassTargetType match {
-                      case Some(targetType) if StringUtils.isNotEmpty(targetType) =>
-                        (element \ "type").asOpt[String] match {
-                          case Some("way") =>
-                            if (targetType != "way") {
-                              targetTypeFailed = true
-                              throw new InvalidException(
-                                "Element type 'way' does not match target type of '" + targetType + "'"
-                              )
-                            }
-                          case Some("relation") =>
-                            if (targetType != "relation") {
-                              targetTypeFailed = true
-                              throw new InvalidException(
-                                "Element type 'relation' does not match target type of '" + targetType + "'"
-                              )
-                            }
-                          case Some("node") =>
-                            if (targetType != "node") {
-                              targetTypeFailed = true
-                              throw new InvalidException(
-                                "Element type 'node' does not match target type of '" + targetType + "'"
-                              )
-                            }
-                          case x =>
+              // parse the results. Overpass has its own format and is not geojson
+              val elements = (payload \ "elements").as[List[JsValue]]
+              try {
+                elements.foreach { element =>
+                  // Verify target type if we are given one.
+                  challenge.creation.overpassTargetType match {
+                    case Some(targetType) if StringUtils.isNotEmpty(targetType) =>
+                      (element \ "type").asOpt[String] match {
+                        case Some("way") =>
+                          if (targetType != "way") {
                             targetTypeFailed = true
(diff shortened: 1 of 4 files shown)


jake-low:
I figured out how to fix the tests, and verified that it's working locally (creating and then rebuilding a challenge from the mr3 UI and confirming that the `/rebuild` endpoint still works). This is ready for review now.

app/org/maproulette/provider/ChallengeProvider.scala
```diff
@@ -32,15 +33,26 @@ class ChallengeProvider @Inject() (
...
   private val logger = LoggerFactory.getLogger(this.getClass)
 
+  /**
+    * Runs a block using a connection borrowed from the dedicated background
+    * pool so that heavy, long-held build operations (e.g. bulk task deletes,
+    * recomputing task priorities) don't draw from the default pool, which can
+    * starve normal HTTP requests and cause 503 errors.
+    */
+  private def withBackgroundPool[T](block: Option[Connection] => T): T =
+    this.backgroundDb.withConnection { c =>
```
CollinBeczak: just a nit: Should this be withTransaction instead of withConnection? mergeUpdate's two writes (task upsert + conditional task_review delete) relied on withMRTransaction opening a real transaction — passing a plain withConnection connection here loses that atomicity.

test/org/maproulette/provider/ChallengeProviderSpec.scala
```diff
@@ -407,8 +431,11 @@ class ChallengeProviderSpec extends PlaySpec with MockitoSugar {
     doAnswer { (invocation: InvocationOnMock) =>
       invocation.getArgument(0).asInstanceOf[Connection => Any](mock[Connection])
     }.when(db).withTransaction(any[Connection => Any])
+    doAnswer { (invocation: InvocationOnMock) =>
+      invocation.getArgument(0).asInstanceOf[Connection => Any](mock[Connection])
+    }.when(backgroundDb).withConnection(any[Connection => Any])
```
CollinBeczak: Mock withTransaction here too, to match the withConnection → withTransaction change above.

CollinBeczak:
everything else lgtm

app/org/maproulette/provider/ChallengeProvider.scala
jake-low: thanks for catching that, yeah I agree.

ljdelight:
This looks good to me 👍 

I did use an AI to review the code, since Fable 5 is somewhat decent at findings. It's a fairly complex finding and concerns the create/rebuild of an overpass challenge that will use the default db pool and the code flows (via task changes) to the background thread pool (where the default db thread will be idle in transact and the background will be doing the work). It seems correct but I didn't actually validate, that's a fairly complex code path to test. Just writing it as an FYI for us in the future.

---

### 1. `buildOverpassQLTasks` still holds a default-pool connection for the entire build loop

(`ChallengeProvider.scala:539` on the PR branch.) The whole Overpass element-processing loop runs inside `this.db.withTransaction`, yet none of the writes inside actually use that connection — `createNewTask` → `mergeUpdate` now borrows from the background pool per task, and the `challengeDAL.update` / `markTasksRefreshed` / `updateBoundingBox` / `updateFinishedStatus` calls at lines 693–709 use their `None` default (default pool, own connections). During a burst of 8 concurrent Overpass rebuilds, 8 default-pool connections sit idle-in-transaction for the full build duration — the largest remaining leak of the exact problem the PR fixes. Suggest switching that outer transaction to `backgroundDb` (or removing it, since it demonstrably isn't providing atomicity for the per-task writes).

jake-low:
If I understand right, the fix is just to remove the explicit withTransaction from buildOverpassQLTasks? I just pushed a commit to do that (untested). [d544439](https://github.com/maproulette/maproulette-backend/pull/1252/commits/d544439513735d971aec577a46b788f7e22457ad); turn off whitespace in the diff viewer (there's a ton of indent formatting noise from this change, but the actual diff is small).

ljdelight:
@jake-low that's my understanding too, without the db override it'll default to the db "default" pool used by the DAL.

LGTM

CollinBeczak:
LGTM

PR #1253, opened 2026-07-06
Add liveness, readiness, and health endpoints

jake-low:
Adds three new health checking endpoints:

- `/liveness` always returns 200 OK (useful to check if the process is alive and answering requests)
- `/readiness` answers 200 OK if the database is reachable, and 503 otherwise (a reasonable proxy for whether the instance is ready to serve traffic)
- `/health` reports on the detailed health of the service. Currently it only checks DB connectivity so the status code should always be the same as `/readiness`, but we can add additional non-critical checks here too (like checking if the osm.org API is online). This endpoint should return a 2xx code if MapRoulette is healthy (even if non-critical external dependencies are down) so we can use it for monitoring / alerting of outages.

Once we deploy this we can switch the uptime monitor to ping `/health` instead of `/service/info` (which is like `/liveness` in that it always succeeds if the process is alive, even if the DB is down) which will help detect outages that we currently miss. The orchestrator can switch to using `/liveness` and/or `/readiness` to see if the process itself is alive and ready to serve requests.

### DIFF
--- a/app/org/maproulette/framework/controller/HealthController.scala
+++ b/app/org/maproulette/framework/controller/HealthController.scala
@@ -1,0 +1,117 @@
+/*
+ * Copyright (C) 2026 MapRoulette contributors (see CONTRIBUTORS.md).
+ * Licensed under the Apache License, Version 2.0 (see LICENSE).
+ */
+package org.maproulette.framework.controller
+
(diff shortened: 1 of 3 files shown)


CollinBeczak (approved):
LGTM

PR #1254, opened 2026-07-12
Add 'paused' state to challenges to prevent task completion and review

CollinBeczak:
This update introduces a 'paused' boolean field to the Challenge model, allowing challenges to be paused. When a challenge is paused, tasks associated with it cannot be completed or reviewed, enforcing a lock on task actions. The database schema is updated accordingly, and relevant service and data access layers have been modified to handle this new state. Tests have been added to ensure the functionality works as intended, preventing task actions while a challenge is paused.

Merge in tandom with: https://github.com/maproulette/maproulette3/pull/2868


jake-low (approved):
I like this as a feature concept; it complements discoverable and archived states but serves a distinct purpose. Code LGTM.

PR #1255, opened 2026-07-13
Refactor task retrieval logic in TaskDAL to improve performance

CollinBeczak:
Replaced the slow SQL query for fetching project details with a more efficient method using the challenge manager. Additionally, optimized cache handling for new tasks to skip unnecessary lookups, enhancing overall performance and reducing latency.

seemingly makes challenge build speed 2x-3x faster. (At least locally goes from 4,000 tasks per 10 seconds to 10,000 tasks per 10 seconds on challenge build)


jake-low (approved):
Reviewed this code as best I can. Looks good from what I can tell (though there are some Scala-isms in here that are above my level of understanding TBH). Nonetheless, feel free to merge.

PR #1268, opened 2026-08-31
Moderated team images for challenge cards

CollinBeczak:
## What

Adds team-owned images that can be displayed on challenge cards, with a
request/approve moderation flow.

A team member uploads an image, which lands as a pending request. A superuser
approves or rejects it. Once approved, any active member of that team can
attach it to their challenges. Deleting an image detaches it from every
challenge that used it, so revoking an image actually clears the cards showing
it.

New endpoints (`conf/v2_route/teamimage.api`):

| Method | Path | Purpose |
| --- | --- | --- |
| POST | `/team/:teamId/image` | Upload an image as a pending request |
| GET | `/team/:teamId/images` | List a team's images |
| GET | `/teamImages/available` | Approved images the caller may use |
| GET | `/teamImages/pending` | Pending requests (superuser) |
| PUT | `/teamImage/:id/approve` | Approve a request (superuser) |
| PUT | `/teamImage/:id/reject` | Reject a request (superuser) |
| DELETE | `/teamImage/:id` | Delete an image |
| GET | `/teamImage/:id/file` | Serve the image bytes |

Challenge json now carries `teamImageId` and a derived `avatarUrl`.

## Also in this PR

Lock fixes that this work depends on:

- **Review claims lock as a bundle.** Claiming a review bundle previously
  wrote one `locked` row per task; it now writes a single row on the primary
  task with the members in `bundled_tasks`, the same shape edit locks use.
  Unclaiming resolves the bundle and clears the claim on every member, and
  leftover review-claim rows from a previous claim are dropped rather than
  left holding tasks locked.
- **`lockBundle` no longer allows overlapping rows.** It folds any of the
  caller's own rows covering a bundle member into the primary row; two rows
  covering the same task would break the `singleOpt` lookups in
  `resolveLockHolder` / `resolveLockBundle`.
- **Paused challenges reject locks.** `startOnTask` and the bundle variant
  return an error when the parent challenge is paused.
- **Conflict responses include `parentId`** next to `parentName`, so the
  client can link to the blocking task's challenge instead of only naming it.

## Migration

`conf/evolutions/default/121.sql` — creates `team_images` and adds
`challenges.team_image_id` (FK, `ON DELETE SET NULL`). Downs drop both.

## Testing

- `TeamImageSpec` — model/status behavior
- `TeamImageRepositorySpec` — repository CRUD and status transitions


jake-low:
Let's discuss this at our next internal sync meeting. I'm not sure we want to continue expanding the "teams" feature in MapRoulette; I view this as a legacy feature that should eventually be replaced by integration with e.g. https://mapping.team/

However, I do definitely think we should have better tools for indicating which organization, company, or group is the owner of a challenge (better accountability for organized editing guidelines, etc) and it sounds like this would be a step in that direction, so maybe it's worth pursuing. Just want to understand the tradeoffs better before we commit to it.

CollinBeczak:
Sounds good, i think that that is a good direction at first thought. Will still probably need most of the teams related changes, but yeah, might just swap out where the teams themselves actually live to there.

